PR #2, opened 2024-11-12
init: repo structure

steveej:
the repository will support multiple different languages

nix
- [x] nix
- [x] nixos host
- [ ] nixos shared host module
- [ ] nixos integration test with shared module

rust
- [ ] binary
- [ ] library

IaC tooling structure or defer?
- [ ] pulumi

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+# Used by https://direnv.net
+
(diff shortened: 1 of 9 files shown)


JettTech (approved):
I agree with the thinking expressed in your prior comment and think we can benefit from having this first iteration with the flake as our shared baseline in the default branch. Merge away!

PR #35, opened 2024-12-10
Initial commit for HAL/inventory crate

mattgeddes:
This has the start of the HAL crate, which includes a binary to output similar JSON to the rector project for Nix. The docs have  a 2 line example showing how to pull the inventory, which the agent will compare against the previous inventory and push back to HQ in the event that it's different from the previous inventory. The crate can also be used to parse the inventory stored in the central database. It gives us a single, consistent view of the hardware and platform. 

There's more that we'll add over time, but this is a start. I'm separating the LED changes out into a separate PR.


steveej (commented):
submitting review items that i did last week. will revisit once more at least once to understand how the JSON informs the host configuration and how hardware-independent that can really be.

steveej (approved):
i did resolve the "merge conflicts" on github. it's _slightly_ different because of spaces. @mattgeddes i'll leave the decision to you whether you want to merge this or close it as the content has already landed.

PR #36, opened 2024-12-10
feat(flake,nix/{hosts,modules}): add hetzner cloud cpx compatible host

steveej:
this adds an opionated NixOS config for Hetzner's CPX11 VPS.


mattgeddes (approved):
Looks fine to me. I know that Hetzner uses KVM under the hood -- do we know which VM attributes they map into the DMI/SMBIOS space inside the VM guest? Specifically thinking about model and serial number info and anything useful they provide in the custom OEM strings section.

PR #38, opened 2024-12-16
Nats Baseline

JettTech:
>Nats PR#0

### Updates:
  - adds  `rust` dir
  - adds nats workload service
  - add nats holo-agent client
  - adds the following nats util-libs:
    - nats leaf-server setup service
    - nats client service
    - adds jetstream microservice
    - adds mongodb schema (and indexing setup) for workloads and hosts
    - adds mongodb interface
  - adds a script to set-up orchestrator servers and Operterator JWT Auth


mattgeddes (commented):
A few comments, but none that ought to hold up merging these changes.

I don't see any main functions or other program entry points below, and we never did any kind of review or spec for this PoC. So it's not obvious how all of these components fit together, nor how they fit in with the architecture and requirements that were defined earlier. When can I see the rest of the code alongside this in order to be able to review it more substantially?

zeeshan595 (approved):

steveej (commented):
nice progress on the structure and aligning on the initial scope :raised_hands: 

i just resolved the merge conflict with main so CI starts running here and you can start iterating on that independently.

even before this is merged i'd like to start implementing the first workload implementation on the host side. for that the minimal scenario that i need to understand how to set up and instrument is:

* minimal leaf node configuration for the holo-agent. i'm okay having that unauthenticated on `::1` or a unix socket if that's supported
* example for how i can send from the `nats` CLI where i can submit a workload request to the holo-agent as if it came from the orchestrator.

i'd greatly appreciate help with that :pray: 

that said, i also don't think we have to block this PR on the workload implementation as it can just remain a TODO in that part of the code.

thanks so far and looking forward to more :pray:

mattgeddes (approved):
Thanks Lisa!

steveej (approved):
great work, thank you :raised_hands: 

i did a few CI related things for/in this PR
* opened a [PR to make postbuildstepper case-insensitive](https://github.com/holochain/holochain-infra/pull/192) so you can run all of CI. it's already deployed though the PR is still to be approved and merged.
* resolved the clippy warnings
* added mongdodb and nats binaries to the rust-workspace test derivation so they can be used within rust tests

TODO (@JettTech): spin up a mongodb for each test instance with a randomly allocated port and either in-memory db or temporary file. we could postpone that into a subsequent PR and ignore/feature-gate the mongodb tests for now.

PR #39, opened 2024-12-17
feat(nats-server): add nix packages, nixos modules and example host configs, holo-agent stub module

steveej:
- [x] depends on #36
- [x] depends on #38


zeeshan595 (approved):
approved

PR #42, opened 2024-12-18
Nats: db proposal

JettTech:
> Nats PR#2

### Updates:
  - adds the following nats util-libs:
    - adds mongodb schema (and indexing setup)
    - adds mongodb interface
  - cleans up baseline
    - removes endpoint files and extracts common call sig into the service `call` method
    - updates the js service endpoint handler to take a generic type <T>
    - standardizes input and output for workload service


rust/util_libs/src/db/schemas.rs
```diff
@@ -37,14 +148,43 @@ impl IntoIndexes for Host {
...
-    pub _id: String, // Mongodb ID
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub _id: Option<MongoDbId>,
     pub version: SemVer,
     pub nix_pkg: String, // (Includes everthing needed to deploy workload - ie: binary & env pkg & deps, etc)
-    pub assigned_hosts: Vec<String>, // Host Device IDs (eg: mac_id)
+    pub assigned_developer: String, // *INDEXED*, Developer Mongodb ID
+    pub min_hosts: u16,
+    pub system_specs: SystemSpecs,
+    pub assigned_hosts: Vec<String>, // Host Device IDs (eg: assigned nats server id)
```
zeeshan595, line 186: This should be mongodb id for hosts

JettTech: We discussed the reasons for and against this and implemented the decision in PR  #51 .

zeeshan595 (approved):

rust/clients/host_agent/src/workload_manager.rs
```diff
@@ -49,34 +50,33 @@ pub async fn run(user_creds_path: &str) -> Result<(), async_nats::Error> {
...
         })
         .await?;
 
     // ==================== DB Setup ====================
-
     // Create a new MongoDB Client and connect it to the cluster
-    // let mongo_uri = get_mongodb_url();
-    // let client_options = ClientOptions::parse(mongo_uri).await?;
-    // let client = MongoDBClient::with_options(client_options)?;
+    let mongo_uri = get_mongodb_url();
```
mattgeddes, line 70: Should we really be calling into MongoDB from the host_agent code? The agent should communicate with the services running centrally, which in turn is the only touch point to the central MongoDB instance. All communication from the agent goes through the API, which is now using NATS as a transport and won't go direct to things like MongoDB.

rust/util_libs/src/db/mongodb.rs
```diff
@@ -139,6 +156,11 @@ where
     }
 }
 
+// Helpers:
+pub fn get_mongodb_url() -> String {
+    std::env::var("MONGO_URI").unwrap_or_else(|_| "mongodb://127.0.0.1:27017".to_string())
```
mattgeddes, line 161: This doesn't seem right either. We won't be running a MongoDB instance on the holoports and other workload hosts.

rust/util_libs/src/db/mongodb.rs
```diff
@@ -245,8 +267,18 @@ mod tests {
 
         fn get_mock_host() -> schemas::Host {
```
mattgeddes, line 268: Is this a placeholder for the inventory? We can probably replace it soon with a call to the hpos_hal::inventory module. The schema itself is still under some change, but that's OK.

mattgeddes (commented):
A few questions/comments about the MongoDB stuff.

rust/clients/host_agent/src/workload_manager.rs
JettTech: You're 100% correct.  This was added here for the convenience of testing in the first iteration.  The host agents will not have a mongodb intance.  I have a PR out that both adds the workload api for the orchestrator and separates out the api for the orchestrator and host such that only the orchestrator service api has access to mongodb.

rust/util_libs/src/db/mongodb.rs
JettTech: Agreed.  See above. :)

rust/util_libs/src/db/mongodb.rs
JettTech: The `Host` struct that this test is referencing is schema for the Host in mongodb.  It is intended to store the relevant inventory data that the `hal::inventory` module generates.  I agree that we could start to align the two structs, even if the `hal::inventory` is subject to change.  I'll take a look into aligning those this week. :+1:

PR #45, opened 2024-12-20
Nats: pw-server edits

JettTech:
### Updates:
 - removes the auth block for local leaf (as I don't think we will need that in our arch)
 - corrects the const for the local leaf server path. (nb: it was missing the `.conf` extension)
 - removes old leaf server conf
 - add new centralized server (hub server) config file for the pw-auth


steveej (approved):
valuable iterative improvement, thanks!

PR #47, opened 2024-12-24
Added first half of agent CLI parsing + rustfmt

mattgeddes:
This keeps the current behaviour (for now) of starting the NATS fork/exec path by default when no arguments are specified, but also adds the beginnings of the CLI stuff raised in https://github.com/Holo-Host/holo-host-private/issues/37. I've plumbed through some trivial output for the `model-info` subcommand. I'll plumb through the network test code as a separate pull request, as it will also include a new crate and that'll be easier to review independently. There's a `daemonize` subcommand that the agent's systemd unit can use to start the agent as a daemon and we should look to move toward that, rather than have the daemon the default path.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -371,6 +371,46 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap"
+version = "4.5.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


zeeshan595 (commented):
approved

steveej (approved):
extending my approval of https://github.com/Holo-Host/holo-host/pull/48 here as that includes this PR as well. i didn't notice this earlier so please bare with my commenting there on content that was introduced here.

PR #50, opened 2025-01-13
feat(holo-agent): improvements and nixos module with integration test

steveej:
- [x] debug infinite recursion in test
- [x] service readiness check for agent


rust/clients/host_agent/src/main.rs
```diff
@@ -36,28 +36,36 @@ async fn main() -> Result<(), AgentCliError> {
...
+            daemonize(daemonize_args).await?;
         }
         Some(agent_cli::CommandScopes::Host { command }) => host_cmds::host_command(command)?,
         Some(agent_cli::CommandScopes::Support { command }) => {
             support_cmds::support_command(command)?
         }
         None => {
             log::warn!("No arguments given. Spawning host agent.");
-            daemonize().await?;
+            daemonize(&Default::default()).await?;
```
mattgeddes, line 49: Can we change this to no longer do this by default, but rather show the command line help if no args are given? Given the discussion around unauthenticated remote control yesterday, I think this opens up a bunch of scope for accidental destruction of machines that don't belong to us.

rust/clients/host_agent/src/workload_manager.rs
```diff
@@ -12,25 +12,29 @@
 */
 
 use anyhow::{anyhow, Result};
+use async_nats::Message;
 use mongodb::{options::ClientOptions, Client as MongoDBClient};
```
mattgeddes, line 16: Why is the agent using the MongoDB crate? The MongoDB instance is central and the host_agent running on remote, distributed nodes. We won't have MongoDB on the holoports. This seems like a mistake.

rust/clients/host_agent/src/gen_leaf_server.rs
```diff
@@ -1,33 +1,54 @@
...
 
-pub async fn run(user_creds_path: &str) {
+pub async fn run(
+    user_creds_path: &Option<PathBuf>,
+    maybe_store_dir: &Option<PathBuf>,
+) -> anyhow::Result<()> {
     let leaf_server_remote_conn_url = nats_server::get_hub_server_url();
     let leaf_client_conn_domain = "127.0.0.1";
-    let leaf_client_conn_port = 4111;
+    let leaf_client_conn_port = std::env::var("NATS_LISTEN_PORT")
```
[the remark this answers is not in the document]
steveej: yes, i'd prefer all of these to be configurable via CLI arguments. i didn't initially plan to make these changes within the scope of this PR, and so i just added the flexibility required for the integration test to be meaningful.

nix/checks/holo-agent-integration-nixos.nix
```diff
@@ -0,0 +1,220 @@
...
+
+        # holo.orchestrator.enable = true;
+        holo.nats-server.enable = true;
+        services.nats.settings = {
+          accounts = {
+            SYS = {
+              users = [
+                {
+                  user = "admin";
+                  "password" = "admin";
```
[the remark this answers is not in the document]
steveej: this is in the context of a an integration tests that spawns VMs on CI with only intra-VM networking, so no connectivity either way into or from the test environment. the VMs are immediately stopped and removed once the test finishes.

rust/clients/host_agent/src/workload_manager.rs
steveej: intuitively i agree. general tech debt cleanup isn't the scope of this PR and since this is pre-existing code i'll defer. still interesting for @JettTech who i expect to have a definitive answer for this question.

rust/clients/host_agent/src/gen_leaf_server.rs
```diff
@@ -1,33 +1,54 @@
...
+        std::fs::create_dir_all(store_dir).context("creating {store_dir:?}")?;
+        (store_dir.clone(), None)
+    } else {
+        let maybe_tempfile = tempdir()?;
+        (maybe_tempfile.path().to_owned(), Some(tempdir))
+    };
 
     let jetstream_config = JetStreamConfig {
-        store_dir: format!("{}/leaf_store", nsc_path),
+        store_dir,
```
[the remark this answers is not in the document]
steveej: agreed. it's many small files from what i've seen so far so SSD would be preferred if available. i'm not conclusive yet where the condition will be determined whether we have SSD storage available. this deserves a separate ticker for sure.

rust/clients/host_agent/src/workload_manager.rs
JettTech: This looks to be the same question a that on pr #42.  I'll share a link to the answer here for a bit less redundancy: https://github.com/Holo-Host/holo-host/pull/42#discussion_r1922884133.

In sum though, this tech debt clean up has already been managed in pr #53.  I think we're all on the same page that only the orchestrator will have access to mongodb. :smiley:

rust/clients/host_agent/src/main.rs
steveej: addressed in via 7156d11298bbf452396b37b874fa584c5464de1e

rust/util_libs/src/nats_server.rs
```diff
@@ -42,6 +66,24 @@ pub struct LeafServer {
     server_handle: Arc<Mutex<Option<Child>>>,
 }
 
+// TODO: consider merging this with the `LeafServer` struct
```
JettTech, line 69: I agree with your comment.  What is the value in having these be separate? It looks like the `NatsConfig` is made up entirely with some of the items in `LeafServer`,  so what do you think about maybe having the config be apart of the LeafServer struct, we could even serde flatten it so that it all prints out to root level:
```
#[derive(Debug, Clone)]
pub struct LeafServer {
    pub config_path: String,
    #[derive(flatten)]
    pub config: NatsConfig,
    server_handle: Arc<Mutex<Option<Child>>>,
}

#[derive(Serialize)]
struct NatsConfig {
    server_name: String,
    host: String,
    port: u16,
    jetstream: JetStreamConfig,
    leafnodes: LeafNodes,
    debug: bool,
    trace: bool,
    logtime: bool,
}
```

Thoughts?

rust/clients/host_agent/src/main.rs
```diff
@@ -36,28 +36,36 @@ async fn main() -> Result<(), AgentCliError> {
...
+    let _ = workload_manager::run(
+        "host_id_placeholder>",
+        &args.nats_leafnode_client_creds_path,
+        args.nats_connect_timeout_secs,
+    )
+    .await?;
+
+    // Only exit program when explicitly requested
+    tokio::signal::ctrl_c().await?;
+
```
JettTech, line 69:
```suggestion
    let host_client = workload_manager::run(
        "host_id_placeholder>",
        &args.nats_leafnode_client_creds_path,
        args.nats_connect_timeout_secs,
    )
    .await?;

    // Only exit program when explicitly requested
    tokio::signal::ctrl_c().await?;
    
    host_client.close().await?;
```

It looks like we're returning the worload manager nats client back here.  Is there a reason e shouldn't be reading it to close the client after the "exit" request is received?

JettTech (approved):
Some really strong updates in this PR.  Nicely done @steveej! :tada: 

Overall, I'm super excited about the addition of this CI setup and the clean-up of the nats leaf server and agent setup.  I left a couple minor comments/suggestions.  I also have one overall question about linting - would it be possible to have a common setup for the linting either as a nix setting or command?  I notice that there are a fair amount of files that only have linting edits.  (Is this using smthg other than `cargo fmt`?)

rust/util_libs/src/nats_server.rs
steveej: agreed. i didn't want to work through that as if would've been an even bigger (unplanned) refactor. my addition was motivated by being able to use `serde_json::to_string` for the config generation instead of manually formatting the config.

steveej:
thanks for the review @JettTech and co-debugging of the integration test the other day :pray: 

> I also have one overall question about linting - would it be possible to have a common setup for the linting either as a nix setting or command? I notice that there are a fair amount of files that only have linting edits. (Is this using smthg other than `cargo fmt`?)

it's indeed possible and already set up in our repo :-) we're using treefmt via treefmt-nix and it can be called from the shell via `nix fmt`. you can learn from in our [nix/formatter.nix](https://github.com/Holo-Host/holo-host/blob/997f9acba217f6f7042264941bb8a9fdbace4a7a/nix/formatter.nix) to learn more

rust/util_libs/src/nats_server.rs
JettTech: Yea, I saw that.  It's very clean now.. love this change!

PR #51, opened 2025-01-16
Workload & Mongodb Schema

zeeshan595:


rust/services/workload/src/lib.rs
```diff
@@ -96,7 +98,7 @@ impl WorkloadApi {
             |workload: schemas::Workload| async move {
                 let workload_query = doc! { "_id":  workload._id.clone() };
                 let updated_workload = to_document(&workload)?;
-                self.workload_collection.update_one_within(workload_query, UpdateModifications::Document(updated_workload)).await?;
+                self.workload_collection.update_one_within(workload_query, UpdateModifications::Document(doc!{ "$set": updated_workload })).await?;
```
zeeshan595, line 101: You need to add this `$set`.

You cannot send the document itself. Mongodb needs a command what to do with the data you are sending it. `$set` will replace the existing data with the data you send.

rust/util_libs/src/db/schemas.rs
```diff
@@ -44,9 +50,15 @@ pub struct RoleInfo {
...
     pub jurisdiction: String,
-    pub roles: Vec<RoleInfo>,
+    pub permisions: Vec<UserPermission>,
+    pub deleted: bool,
+    pub deleted_at: Option<DateTime>,
+    pub updated_at: Option<DateTime>,
+    pub created_at: Option<DateTime>,
+
+    pub email: String, // we need to remove this from the collection (personal info)
+    pub roles: Vec<RoleInfo>, // I suggest we remove this. Use the mongodb query to map the role when we fetch the user data
```
zeeshan595, line 61: `email` is considered personal information. We should have another collection and we move the email of the user in that collection. If we need to track the user I suggest we use a `user_name` or something else

The role info. I suggest we remove that. It does not fit with the pattern that we have.
An alternative that I would be okay with is creating 2 different fields 
```rust
pub struct User {
  pub developers: vec<MongoDbId>,
  pub hosters: vec<MongoDbId>,
}
```

rust/util_libs/src/db/schemas.rs
```diff
@@ -82,7 +94,7 @@ pub struct Developer {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub _id: Option<MongoDbId>,
     pub user_id: String, // MongoDB ID ref to `user._id` (which stores the hoster's pubkey, jurisdiction and email)
-    pub requested_workloads: Vec<String>, // MongoDB ID refs to `workload._id`
+    pub active_workloads: Vec<MongoDbId>, // MongoDB ID refs to `workload._id`
```
zeeshan595, line 97: We should use MongodbId not strings

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +235,84 @@ impl WorkloadApi {
...
+        let mut workload: schemas::Workload = serde_json::from_slice(&payload_buf)?;
+        log::trace!("New workload to assign. Workload={:#?}", workload.clone());
+
+        // 1. fetch existing workload document
+        match self.workload_collection.get_one_from(doc! { "_id":  workload._id.clone() }).await?{
+            Some(workload_doc) => {
+                // 2. remove workload from hosts
+                self.host_collection.update_one_within(
+                    doc!{ "_id": { "$in": workload_doc.assigned_hosts } },
+                    UpdateModifications::Document(doc!{ "$pull": { "assigned_workloads": workload_doc._id.clone() } })
```
JettTech, line 268: Can we check to see whether the host has the capacity to support the updated workload requirements prior to removing removing the workload?

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +235,84 @@ impl WorkloadApi {
...
     // NB: Automatically published by the nats-db-connector
     pub async fn handle_db_update(&self, msg: Arc<Message>) -> Result<types::ApiResult, anyhow::Error> {
         log::debug!("Incoming message for 'WORKLOAD.update'");
         
         let payload_buf = msg.payload.to_vec();
-        let workload: schemas::Workload = serde_json::from_slice(&payload_buf)?;
-        log::trace!("New workload to assign. Workload={:#?}", workload);
+
+        let mut workload: schemas::Workload = serde_json::from_slice(&payload_buf)?;
+        log::trace!("New workload to assign. Workload={:#?}", workload.clone());
```
JettTech, line 260:
```suggestion
        log::trace!("Workload to update. Workload={:#?}", workload.clone());
```
copy/paste typo

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +235,84 @@ impl WorkloadApi {
...
+            doc!{ "_id": { "$in": workload.clone().assigned_hosts } },
+            UpdateModifications::Document(doc!{ "$push": { "assigned_workloads": workload._id.clone() } })
+        ).await?;
         
-        // TODO: ...handle the use case for the update entry change stream 
+        // 5. add workload to developers
+        self.developer_collection.update_one_within(
+            doc!{ "_id": { "$in": workload.clone().assigned_developer } },
+            UpdateModifications::Document(doc!{ "$push": { "assigned_workloads": workload._id.clone() } })
+        ).await?;
```
JettTech, line 292:
```suggestion
```
I recommend removing this call as well as the call to remove the workload from the developer collection.

If we don't remove the workload reference from the developer collection, we also wouldn't need to add it back to the developer collection.

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +235,84 @@ impl WorkloadApi {
...
+                // 3. remove workload from developers
+                self.developer_collection.update_one_within(
+                    doc!{ "_id": { "$in": workload_doc.assigned_developer } },
+                    UpdateModifications::Document(doc!{ "$pull": { "assigned_workloads": workload_doc._id.clone() } })
+                ).await?;
+            },
+            None => {
+                log::warn!("Workload not found. Unable to remove workload.");
+            }
+        }
```
JettTech, line 280:
```suggestion
```
Why do we need to remove the doc from the developer collection, if it is only being updated and the ID remains the same?

I recommend removing this call as the ID of the workload will remain unchanged during an call to update.

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +235,84 @@ impl WorkloadApi {
...
+
+        // 4. add workload to hosts
+        self.host_collection.update_one_within(
+            doc!{ "_id": { "$in": workload.clone().assigned_hosts } },
+            UpdateModifications::Document(doc!{ "$push": { "assigned_workloads": workload._id.clone() } })
+        ).await?;
```
JettTech, line 286: s a follow-up to the host collection change request above, can we only make this call to assign a new host whenever the current host does *not* have the required capacity to host the updated workload?

Also, this call should only assign a host based on capacity.

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +235,84 @@ impl WorkloadApi {
...
+            doc!{ "_id": { "$in": workload.clone().assigned_developer } },
+            UpdateModifications::Document(doc!{ "$push": { "assigned_workloads": workload._id.clone() } })
+        ).await?;
+
+        workload.updated_at = Some(DateTime::from_chrono(chrono::Utc::now()));
+        let updated_workload_doc = to_document(&workload)?;
+        self.workload_collection.update_one_within(
+            doc! { "_id": workload._id.clone() },
+            UpdateModifications::Document(doc!{ "$set": updated_workload_doc }),
+        ).await?;
```
JettTech, line 299: The workload has already been updated, which is the trigger for this function to be called.
```suggestion
```
The workload has already been updated, as that is the trigger for this function to be called.  Calling this again would create an undesirable loop.

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +235,84 @@ impl WorkloadApi {
...
         
-        // TODO: ...handle the use case for the update entry change stream 
+        // 5. add workload to developers
+        self.developer_collection.update_one_within(
+            doc!{ "_id": { "$in": workload.clone().assigned_developer } },
+            UpdateModifications::Document(doc!{ "$push": { "assigned_workloads": workload._id.clone() } })
+        ).await?;
+
+        workload.updated_at = Some(DateTime::from_chrono(chrono::Utc::now()));
```
JettTech, line 294: Let's move this to where the workload is being updated in the `update_workload` method above (..I think it's on line 105).

PS. I love this pattern of adding `updated_at` and `deleted_at`.  :)

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +235,84 @@ impl WorkloadApi {
...
+        ).await?;
 
         let success_status = WorkloadStatus {
-            id: workload._id,
+            id: match workload._id.clone() {
+                Some(oid) => Some(oid.to_hex()),
+                None => None
+            },
             desired: WorkloadState::Running,
             actual: WorkloadState::Running,
```
JettTech, line 307:
```suggestion
            desired: WorkloadState::Updating,
            actual: WorkloadState::Updating,
```
Can we add a state to represent the state wherein a workload will need the host to update a workload?

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +235,84 @@ impl WorkloadApi {
...
         let payload_buf = msg.payload.to_vec();
-        let workload: schemas::Workload = serde_json::from_slice(&payload_buf)?;
-        log::trace!("New workload to assign. Workload={:#?}", workload);
+
+        let mut workload: schemas::Workload = serde_json::from_slice(&payload_buf)?;
+        log::trace!("New workload to assign. Workload={:#?}", workload.clone());
+
+        // 1. fetch existing workload document
+        match self.workload_collection.get_one_from(doc! { "_id":  workload._id.clone() }).await?{
+            Some(workload_doc) => {
```
JettTech, line 264: The `Message` payload should serialize into the `Workload` struct already, so we should be able to avoid making this fetch call and use the `assigned_hosts` field directly.
```suggestion
```

rust/services/workload/src/lib.rs
```diff
@@ -263,7 +322,10 @@ impl WorkloadApi {
         // TODO: ...handle the use case for the delete entry change stream
```
JettTech, line 322: Let me know if you have any Qs and/or want to get together to discuss the use case.

JettTech (changes requested):
I noticed that the delete fn is still unimplemented, so some of the feedback may be a bit premature.  That said, I went ahead and finished the review since we discussed it via DM.  Also, I do think that this is a great start!  :muscle: 

I really appreciate the clean up of the mongodb id type and the addition of the `updated_at` and `deleted_at` fields.  I also responded to some of the questions re: the `User` collection and left some feedback on the update method; let me know if you have any Qs you'd like to sync on.

rust/util_libs/src/db/schemas.rs
zeeshan595: Makes sense. I just assumed you could have that because with the old way you could do that :D

rust/services/workload/src/lib.rs
zeeshan595: This is because the hosts and developers could be changed and the workload from the payload will have the newer assigned hosts and developers.

Which is why it pulls the workload from existing once and then pushes it again afterwards

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +241,60 @@ impl WorkloadApi {
...
         
         Ok(types::ApiResult(success_status, None))
     }
-
-    // Zeeshan to take a look:
-    // NB: Automatically published by the nats-db-connector
+  
+    // !!!!!!!!!!!!!!DEPRECATED?!!!!!!!!!!!!!!!!!!
+    // NB: This is the stream that is automatically published by the nats-db-connector
+    // triggers on mongodb [workload] collection (delete)
```
JettTech, line 295: Clean up comments, and delete the deprecated function. :)

rust/services/workload/src/lib.rs
```diff
@@ -94,13 +96,20 @@ impl WorkloadApi {
             msg,
             WorkloadState::Running,
             |workload: schemas::Workload| async move {
-                let workload_query = doc! { "_id":  workload._id.clone() };
-                let updated_workload = to_document(&workload)?;
-                self.workload_collection.update_one_within(workload_query, UpdateModifications::Document(updated_workload)).await?;
+                let workload_query = doc! { "_id":  workload._id };
+
+                // update workload updated_at
+                let mut workload_doc = workload.clone();
```
JettTech, line 102: What do you think about automating this and the other metadata changes?  I'm thinking we can just move the insertion/editing of the metadata fields to their associated methods in the `MongodbAPI` struct?  This way the changes would be automated and we wouldn't have to write out the same changes each time.

rust/services/workload/src/lib.rs
```diff
@@ -193,7 +215,7 @@ impl WorkloadApi {
                 let host_id = host._id.to_owned().unwrap();
                 
                 // 4. Update the Workload Collection with the assigned Host ID
-                let workload_query = doc! { "_id":  workload_id.clone() };
```
JettTech, line 196: Yay! So happy to see these clones disappear! :D

rust/services/workload/src/lib.rs
JettTech: Also... what is the state of the deletion handling?

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +241,60 @@ impl WorkloadApi {
...
-        // TODO: ...handle the use case for the update entry change stream 
+        log::trace!("Workload to update. Workload={:#?}", workload.clone());
+
+        // 1. remove workload from all hosts
+        self.host_collection.update_one_within(
+            doc!{},
+            UpdateModifications::Document(doc!{ "$pull": { "assigned_workloads": workload._id } })
+        ).await?;
+        log::info!("Removed workload from previous hosts. Workload={:#?}", workload._id);
+
```
JettTech, line 275: Before assigning the new hosts in the DB, can we add some logic to ensure that the listed hosts still have a matching capacity? ...and they don't then enter into the loop wherein we get all hosts that fit the capacity requirements filter?

rust/services/workload/src/lib.rs
```diff
@@ -219,41 +241,60 @@ impl WorkloadApi {
...
-        log::trace!("New workload to assign. Workload={:#?}", workload);
-        
-        // TODO: ...handle the use case for the update entry change stream 
+        log::trace!("Workload to update. Workload={:#?}", workload.clone());
+
+        // 1. remove workload from all hosts
+        self.host_collection.update_one_within(
+            doc!{},
+            UpdateModifications::Document(doc!{ "$pull": { "assigned_workloads": workload._id } })
+        ).await?;
```
JettTech, line 273: Can we first run a get call for all hosts with the matching assigned workloads and then contrast that list with the updated assigned list?  This way we can only remove the workload ids from the hosts that were removed... and more importantly we will know which hosts will have updated workload and which ones totally new workloads to install and can report that state accordingly in the response.

JettTech (commented):
Clean additions and changes! :muscle:  I'm liking the direction this is taking.  I left some feedback on the mongodb schema and left a request to add some additional logic to the handle_db_update flow.  I also proposed automating the metadata changes by moving them over the mongodb struct itself. (...and I love that new metadata structure, btw. :))

Let me know if you have any Qs. Happy to talk through any/all of this too.

rust/services/workload/src/lib.rs
zeeshan595: I think for now we can just leave them flagged as deleted later on we can add a cron job or some other service to pick them up and move them to an archive database or remove them from the database.

Usually there is also a data retention policy which means in most countries you need to keep the data for a specific period of time before it is removed. So I think it needs another ticket and some thought to build around that.

rust/util_libs/src/db/mongodb.rs
zeeshan595, line 1: I don't like this pattern of creating wrappers around mongodb functions. It is making more work by adding boilerplate code and also some of the more complex functions in mongodb will require a lot of work to be handled such as `aggregate`.

Also, I tried doing `find().skip().limit().toArray()` which is not possible with this setup.

I would prefer just making the collection a pub and not using our own wrapper functions.

zeeshan595: I've manage to create a solution for this by adding the aggregate function

JettTech: Let's look at this together.  The purpose/goal of the mongodb interface struct is actualy to abstract away the boilerplate code from as well as enforce consistent patterns around access and use.  It also allows for clean management of the mongodb pool, which is created under the hood in the rust driver and offers the option for easy cache handling.

I'd love to learn more about the areas you're seeing the addition of boilerplate code. Also, if we have a need for aggregated pipeline, why not just add the function to the mongodb interface itself?  That way other areas of the code can also leverage that same pattern without having to rewrite it everytime.  

In the meanwhile, I'll checkout the `find()...` example you provided. Thanks for sharing that. :)

rust/services/workload/src/lib.rs
JettTech: Totally agree that we should allocate a specific ticket for the deletion clean-up. :)

We still do need to handle the workload deletion case, in that the workload needs to be removed from the associated host collections and marked as deleted in the developer collections.  The workload status also needs to be set to deleted, so that the hosting agent knows to uninstall the workload.

rust/util_libs/src/db/schemas.rs
```diff
@@ -25,52 +25,93 @@ pub use String as DeveloperJWT;
...
 // Providetype Alias for MongoDB ID (mongo's automated id)
-pub use String as MongoDbId;
+pub use bson::oid::ObjectId as MongoDbId;
 
 // ==================== User Schema ====================
 #[derive(Serialize, Deserialize, Clone, Debug)]
-pub enum Role {
-    Developer(DeveloperJWT), // jwt string
-    Host(HosterPubKey),      // host pubkey
+pub enum UserPermission {
```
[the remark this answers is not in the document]
zeeshan595: User's will generally require permissions. For example hosters should be able to access their own hosts and information about it. Developers should be able to access workloads they have deployed e.t.c.

This should be done through assigning permissions to them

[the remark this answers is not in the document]
JettTech: I see an overlap between this and the nats permissions use case.  I understand that, prior to nats, we used to manage permissions in this way, but I think we no longer need to in our new architecture.  That said, I'm open to being persuaded otherwise. :)

JettTech (approved):
I approve... with the small reminder/acknowledgement that the orchestrator code here is a placeholder and there are a few code debt items we expect to resolve in later prs.  Also, let's make sure the CI passes prior to merging. :+1:

Overall, I'm excited to see this move into main, this includes some updates to the mongodb interface and handling!  Thanks for the hard work. :muscle:

JettTech:
@steveej - can you take a look at the above CI failure.  It is fialing with the error: `aarch64-linux.pkgs-rust-workspace-doc was failed because it has failed previously and its failure has been cached.`
https://buildbot-nix-0.infra.holochain.org/#/builders/414/builds/206

steveej:
> @steveej - can you take a look at the above CI failure. It is failing with the error: `aarch64-linux.pkgs-rust-workspace-doc was failed because it has failed previously and its failure has been cached.` https://buildbot-nix-0.infra.holochain.org/#/builders/414/builds/206

i pushed d96abc1f6346691a72f656c640aac009f3348eb7 to avoid the build-time download requirement entirely.

zeeshan595:
Moved to https://github.com/Holo-Host/holo-host/pull/63

PR #52, opened 2025-01-17
Added first pass of netdiag

mattgeddes:
This does a DNS lookup, HTTP request, with TLS, from scratch, with specific diagnostics when things fail (unlike when higher level interfaces might be used). We can later extend this to also use the authentication interface when that's implemented, and probably also extend it to make raw connections to services within the NATS infrastructure. That can all come later. 

We'll also likely call this from some of the pre-migration checks that are planned.


rust/netdiag/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+}
+
+pub fn do_requests(cfg: &NetDiagConfig) -> Vec<RequestStats> {
+    let mut ret: Vec<RequestStats> = vec![];
+    for query in &cfg.queries {
+        debug!(
+            "Querying: {}:{} via DNS server {}",
+            query.hostname, query.port, query.nameserver
+        );
+        // TODO: do_request() should always return a RequestStats object -- even on failure.
```
steveej, line 221: looks like this TODO is resolved now

rust/netdiag/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+            stats.result = Err(NetDiagError::IOError(e));
+            stats.finalise();
+            return stats;
+        }
+    };
+    stats.end_phase();
+
+    let host = format!("{}:{}", ip, cfg.port);
+    stats.start_phase(PhaseType::TcpConnect);
+    // TODO: replace all unwraps and expects, and bubble it all upward
```
steveej, line 251: looks leftover

steveej (commented):
what are the uses cases for this?

one i can think of is to run this from the agent when the local NATS server fails to establish an outgoing leafnode connection?

i'm not sure how likely it is that the outgoing NATS connection wouldn't work and however we want to transport the result from such a test would :thinking: is result submission planned for this?

final note: do you think it's worth writing a small integration test for this against a test VM that runs DNS and HTTPS?

pandeyAK:
@mattgeddes Which task/ticket is this PR related to?

mattgeddes:
> what are the uses cases for this?

There are a few. Firstly, it's a tool that Rob and the docs can have hosters potentially track down network issues. It's also something that can be called from the selfcheck code we want to implement to have a node validate its state before we try to migrate it to the new network.

> one i can think of is to run this from the agent when the local NATS server fails to establish an outgoing leafnode connection?

That's another potential case. We want to potentially extend this to also check that we can authenticate when a crate exists for that, and to add a NATS client to see that we can establish a NATS session. It would all be the same crate, but called from various places to allow the interactive and non-interactive cases. I think there's even comments to this effect in part of the code.

> i'm not sure how likely it is that the outgoing NATS connection wouldn't work and however we want to transport the result from such a test would 🤔 is result submission planned for this?
> 

Yes. Once we have authentication, we'd likely add a call for that and one or more NATS calls to validate various layers of that. Much of this code pre-dates the NATS POC and subsequent non-design/non-scoping. It also made for a good first pass at something to fill in some of the CLI stuff while you folks were all on leave and other tasks were subsequently blocked.

> final note: do you think it's worth writing a small integration test for this against a test VM that runs DNS and HTTPS?

 More than happy to. If you can point me at an example I can use as a place to start, and point me to the docs for how to execute the tests, I'll definitely raise a task and separate PR for that.

mattgeddes:
> @mattgeddes Which task/ticket is this PR related to?

It's an implementation to fill in one of the placeholder subcommands in https://github.com/orgs/Holo-Host/projects/10?pane=issue&itemId=84171223&issue=Holo-Host%7Cholo-host-private%7C37. It mainly uses code that was written last year and was something I could implement while I was blocked and everyone was off for the holidays.

mattgeddes:
@steveej can I get approval on this? Also see above question/request about Nix and testing from earlier.

steveej (approved):
> can I get approval on this?

:+1: given we adhere to CI feedback i see no counter-indication to merging.



> More than happy to. If you can point me at an example I can use as a place to start, and point me to the docs for how to execute the tests, I'll definitely raise a task and separate PR for that.

> Also see above question/request about Nix and testing from earlier.

in this codebase we have the [holo-agent integration test](https://github.com/Holo-Host/holo-host/blob/4b846967db7244d2699918177a87e95d1e8a4325/nix/checks/holo-agent-integration-nixos.nix) as the only precedence so far. it does show a multi-host setup so the bare-bones of it can serve as a template for this use-case. we'd need two hosts, one that runs the DNS and HTTP server and the other one that runs netdiag.

there's an [overwhelming amount of nixos tests upstream](https://github.com/NixOS/nixpkgs/tree/master/nixos/tests) where we can get snippets from various places to put together a small test for netdiag. e.g. [here's a snippet that enables dnsmasq](https://github.com/NixOS/nixpkgs/blob/667d42c00d566e091e6b9a19b365099315d0e611/nixos/tests/schleuder.nix#L88-L96). here's a very small [test that shows using curl and ffsend](https://github.com/NixOS/nixpkgs/blob/667d42c00d566e091e6b9a19b365099315d0e611/nixos/tests/schleuder.nix#L88-L96) in the testscript against localhost.

i don't expect the test to land with this PR. tracking it in a follow-up ticket as a reminder would be good.

i started a document about interesting nix education resources for our use-cases that i'll share out-of-band.

PR #53, opened 2025-01-20
Orchestrator Workload Client

JettTech:
Updates:
  - Adds orchestrator workload client


pandeyAK:
@JettTech Which task is this PR related to?

rust/clients/orchestrator/src/workloads.rs
```diff
@@ -0,0 +1,182 @@
...
+    // NB: These subjects are published by external Developer (via external api), the Nats-DB-Connector, or the Hosting Agent
+    let workload_service = orchestrator_workload_client
+        .get_js_service(WORKLOAD_SRV_NAME.to_string())
+        .await
+        .ok_or(anyhow!(
+            "Failed to locate Workload Service. Unable to spin up Orchestrator Workload Client."
+        ))?;
+
+    // Published by Developer
+    workload_service
```
zeeshan595, line 83: Seems very similar to the code here
https://github.com/Holo-Host/holo-host/pull/53/files#diff-60145ac38000440a8a0cd1d711c3d6fcc12702bb2fa4c5fff8ac93efd78bc6fcR85
Is there a way we can have a shared function to set these up?

zeeshan595 (approved):

PR #56, opened 2025-01-29
spike to run hivello in a nixos-container (on real hardware)

steveej:
spike(nix/packages): add hivello extra-container

using the previuosly introduced package, provides nix-ld with its
dependencies installed and runs and exposes Xvnc server to provide a
graphical environment for its exectuion.

---

spike(nix/packages): add hivello

ended up relying on nix-ld because i couldn't figure out exactly which
files needed patching or wrapping in which way.

---

part of https://github.com/holo-host/holo-host-private/issues/94

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -306,13 +306,45 @@
         "type": "github"
       }
     },
-    "nixpkgs-unstable": {
+    "nixpkgs-2405": {
       "locked": {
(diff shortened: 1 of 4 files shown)


mattgeddes (approved):
thanks for this. I think I grok where the hivello specific stuff starts and the generic stuff stops.

PR #59, opened 2025-02-06
Feature/holo 98 enforce ticket linking

zeeshan595:
- Adds a new script that enforces new branch rules locally
- Adds a new pipeline to enforce branch names. It will fail if the branch name is not set correctly


zeeshan595:
No longer required

PR #60, opened 2025-02-06
Feature/98 enforce linking to issue

zeeshan595:
Closes #98
I think the project is not assigned properly to this repo which is why the ticket number is not linking up with the issue


.github/hooks/pre-commit
```diff
@@ -0,0 +1,16 @@
...
+
+# Get the current branch name
+branch_name=$(git rev-parse --abbrev-ref HEAD)
+
+# Define the allowed branch name pattern (GitHub issues use numbers)
+pattern='^(feature|bugfix|hotfix|release)\/[0-9]+-[a-z0-9-]+$'
+
+if [[ ! $branch_name =~ $pattern ]]; then
+    echo "❌ Error: Invalid branch name '$branch_name'"
+    echo "✅ Expected format: feature/123-short-description"
```
JettTech, line 11: I actually really like the `feature/...` pattern, but do remember a conversation happening several months ago between Robbie, Mateusz, and Alastair wherein there was an issue using the forward-slash in the branch name.  Namely conflicts with remote branches can arise ...and I think CI/CD tools have issues with this (but that shouldn't affect our current setup).

Are we ok with taking on the risks/issues that were raised at that time?

.github/workflows/enforce-branch-name.yml
```diff
@@ -0,0 +1,35 @@
+name: Enforce Branch Naming
```
JettTech, line 1: I have one question here.  Do we currently have a way to bypass these rules?  If not, what is your approach around the timing of merging in this rule with regard to the currently outstanding PRs that do not follow this rule?  

Two approaches come to mind:
  - we either wait to merge in this rule until those are merged, or
  - we just allow our dev-group to have bypass permissions while these older branches remain outstanding.  

My preference would probably be for the latter, so we can merge this through and have the checks at least coming up.  What are your thoughts?

JettTech (commented):
Looks good overall.  Thanks for this addition!  I left one question about the pattern for our enforced branch names that I think we should discuss/agree on as a team. 

I also left a general question about the approach/timing for merging this with consideration to the currently outstanding branches && PRs that do not follow this naming convention.

.github/workflows/enforce-branch-name.yml
zeeshan595: So any branch that does not follow this rule can create a new branch based from the existing with the correct naming scheme and then create another PR.

I don't think there should be a way to bypass the rule because that is the whole point of setting up a rule in the first place. For existing PRs that conflict they can be merged before merging this PR or you can create a new PR with the same work.

.github/hooks/pre-commit
zeeshan595: I have personally not come across any issues with using `/` in the branch name. The only thing I know of is that you cannot start the branch name with a `.` if you have a slash

e.g. `feature/.lock` is not allowed

.github/workflows/enforce-branch-name.yml
JettTech: Ok. Sounds good.  Let's hold off on merging this then until at least 51 and 53 are merged. :+1:

zeeshan595:
Moved to https://github.com/Holo-Host/holo-host/pull/61

PR #62, opened 2025-02-11
feature: add nix auth test

JettTech:


nix/modules/nixos/holo-orchestrator.nix
```diff
@@ -0,0 +1,177 @@
...
+  ...
+}:
+
+let
+  cfg = config.holo.orchestrator;
+in
+{
+  imports = [
+    inputs.extra-container.nixosModules.default
+  ];
```
steveej, line 22:
```suggestion
```

not necessary in the orchestrator as the orchestrator machines aren't expected to run workloads.

nix/modules/nixos/holo-host-agent.nix
```diff
@@ -36,6 +37,11 @@ in
       default = inputs.self.packages.${pkgs.stdenv.system}.rust-workspace;
     };
 
+    hostAuthScriptPath =  lib.mkOption {
+      type = lib.types.path;
+      default = "${cfg.package}/../../../scripts/hosting_agent_setup ";
```
steveej, line 42: this won't work because `${cfg.package}` evaluates `/nix/store/hbsg3zfgbzns6ak7f29wdn23angw1501-holo-host-workspace-0.1.0`. following `../../../` from there leaves you at `/`. you can use `${flake}` as the top-level directory for the git repository and construct paths with that suffix, e.g. 

```suggestion
      default = "${flake}/scripts/hosting_agent_setup.sh"";
```

(untested, however i added the `.sh` as i believe that's what you wanted to point to)

nix/modules/nixos/holo-host-agent.nix
```diff
@@ -96,6 +137,12 @@ in
         {
           RUST_LOG = cfg.rust.log;
           RUST_BACKTRACE = cfg.rust.backtrace;
+          NSC_PATH = cfg.nats.nscPath;
+          LOCAL_CREDS_PATH = cfg.nats.localCredsPath;
+          HOSTING_AGENT_HOST_NKEY_PATH = cfg.nats.hostNkeyPath;
+          HOSTING_AGENT_SYS_NKEY_PATH = cfg.nats.sysNkeyPath;
+          HPOS_CONFIG_PATH = cfg.nats.hposCredsPath;
+          DEVICE_SEED_DEFAULT_PASSWORD = builtins.toString cfg.nats.hposCredsPw;
```
steveej, line 145: this is a credential and referring to it like this will put in the nix store and thus make it world-readable. we need to find a different solution for this.

```suggestion
          DEVICE_SEED_DEFAULT_PASSWORD_FILE = builtins.toString cfg.nats.hposCredsPwFile;
```

nix/modules/nixos/holo-host-agent.nix
```diff
@@ -64,6 +70,41 @@ in
...
+
+      hposCredsPath = lib.mkOption {
+        type = lib.types.path;
+        default = "/var/lib/server-key-config.json";
+      };
+
+      hposCredsPw = lib.mkOption {
+        type = lib.types.str;
+        default = "pass";
+      };
```
steveej, line 106:
```suggestion
      hposCredsPwFile = lib.mkOption {
        type = lib.types.path;
        default = "/var/lib/holo-host-agent/hpos_creds_pw";
      };
```

the path could be somewhere else. my intention here is to not hardcode some default password. for placing this file accordingly we'll need a different mechanism.

nix/modules/nixos/holo-host-agent.nix
```diff
@@ -106,6 +153,15 @@ in
         pkgs.nats-server
       ];
 
+      preStart = ''
+        init_host_auth_guard() {
+          ${cfg.hostAuthScriptPath} ${builtins.toString cfg.nats.nscPath} ${builtins.toString cfg.nats.sharedCredsPath}
+        }
+        init_host_auth_guard
+        echo "Finished Host Auth Guard Setup"
+        sleep 1 # wait
```
steveej, line 162: what is it waiting for here? can we loop and wait for a well-known state instead of fixed time?

nix/modules/nixos/holo-orchestrator.nix
```diff
@@ -0,0 +1,177 @@
...
+        default = 27017;
+      };
+
+      url = lib.mkOption {
+        type = lib.types.str;
+        default = "${cfg.mongo.listenHost}:${builtins.toString cfg.mongo.listenPort}";
+      };
+
+      services.mongodb = {
+        enable = true;
```
steveej, line 76: i don't see this option referenced anywhere. did you intend that this enables the mongodb service on the nixos machine? if so you can move it to the attribute set at `config = ` further down, which will then configure the system accordingly when this module is included.

nix/modules/nixos/holo-orchestrator.nix
```diff
@@ -0,0 +1,177 @@
...
+        bind_ip = cfg.mongo.listenHost;
+      };
+
+      systemd.services.mongodb.serviceConfig = {
+        LimitNOFILE = 500000;
+      };
+
+      networking.firewall.allowedTCPPorts = [ cfg.mongo.listenPort ];
```
steveej, line 84: this causes the mongodb instance to be reachable by external hosts. is that intentional? i'll defensively suggest against this :relaxed: 
```suggestion
```

nix/modules/nixos/holo-orchestrator.nix
```diff
@@ -0,0 +1,177 @@
...
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    systemd.services.holo-orchestrator = {
+      enable = true;
+
+      after = [
+        "network.target"
+        "network-online.target"
```
steveej, line 140: one of my recent leanings is that we only need one of these. i guess we only need the network interfaces to be up, hence:

```suggestion
        "network.target"
```

nix/modules/nixos/holo-orchestrator.nix
```diff
@@ -0,0 +1,177 @@
...
+          NATS_URL = cfg.nats.url;
+        };
+
+      path = [
+        pkgs.nats-server
+      ];
+
+      preStart = ''
+        init_hub_auth() {
+          ${pkgs.bash}/bin/bash ${builtins.toString cfg.hubAuthScriptPath} ${cfg.nats.listenHost} ${builtins.toString cfg.nats.listenPort} ${builtins.toString cfg.nats.sharedCredsPath} ${builtins.toString cfg.nats.localCredsPath}
```
steveej, line 165:
```suggestion
          ${cfg.hubAuthScriptPath} ${cfg.nats.listenHost} ${builtins.toString cfg.nats.listenPort} ${builtins.toString cfg.nats.sharedCredsPath} ${builtins.toString cfg.nats.localCredsPath}
```

nix/modules/nixos/holo-orchestrator.nix
```diff
@@ -0,0 +1,177 @@
...
+          ORCHESTRATOR_SIGNING_AUTH_NKEY_PATH = cfg.nats.signingAuthNkeyPath;
+          NATS_LISTEN_PORT = builtins.toString cfg.nats.listenPort;
+        }
+        // lib.attrsets.optionalAttrs (cfg.nats.url != null) {
+          NATS_URL = cfg.nats.url;
+        };
+
+      path = [
+        pkgs.nats-server
+      ];
```
steveej, line 161:
```suggestion
        pkgs.bash
      ];
```

steveej (commented):
i'm happy you've brought on this code to work through!

first review pass. if in doubt with any of my comments fallback to anything that helps you get the tests pass on CI, specifically `checks.x86_64-linux.auth-integration-nixos` which was last run here: https://buildbot-nix-0.infra.holochain.org/#/builders/461

nix/modules/nixos/holo-host-agent.nix
JettTech: Awesome, yes!  This is exactly what I was wanting to discover and do!  Thank you. :muscle: :tada:

nix/modules/nixos/holo-host-agent.nix
JettTech: I actually don't think we'll need it  anymore. I'll remove it.

nix/modules/nixos/holo-orchestrator.nix
JettTech: Done.  Thanks for the helpful directions!

nix/modules/nixos/holo-orchestrator.nix
```diff
@@ -0,0 +1,173 @@
...
+
+      signingAuthNkeyPath = lib.mkOption {
+        type = lib.types.path;
+        default = "${cfg.nats.localCredsPath}/AUTH_SK.nk";
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    systemd.services.mongodb = {
```
steveej, line 119: assuming you want to enable the existing nixpkgs module for mondodb, please look at these: https://search.nixos.org/options?channel=24.11&from=0&size=50&sort=relevance&type=packages&query=services.mongodb

```suggestion
    services.mongodb = {
```

nix/modules/nixos/holo-orchestrator.nix
```diff
@@ -0,0 +1,173 @@
...
+      enable = true;
+      bind_ip = cfg.mongo.url;
+      documentation.enable = false;
+      serviceConfig = {
+        LimitNOFILE = 500000;
+      }
```
steveej, line 125: and then these can't be in here with the others as this is indeed for `systemd.services.mongodb`. and is it really needed?

nix/modules/nixos/holo-orchestrator.nix
```diff
@@ -0,0 +1,173 @@
...
+        default = "${cfg.nats.localCredsPath}/AUTH_SK.nk";
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    systemd.services.mongodb = {
+      enable = true;
+      bind_ip = cfg.mongo.url;
+      documentation.enable = false;
```
steveej, line 122: this doesn't exist in the module linked above.

nix/checks/auth-integration-nixos.nix
```diff
@@ -0,0 +1,176 @@
...
+
+    defaults.networking.hosts = {
+      hubIP = [ "${nodes.hub.networking.fqdn}" ];
+    };
+
+    nodes.hub =
+      { ... }:
+      {
+        imports = [
+          flake.nixosModules.holo-nats-server
```
steveej, line 27: this may not be needed depending on whether holo-orchestrator spawns NATS itself or not.

nix/checks/auth-integration-nixos.nix
```diff
@@ -0,0 +1,176 @@
...
+          trace = false;
+          logtime = false;
+        };
+      };
+
+    nodes.host =
+      { ... }:
+      {
+        imports = [
+          flake.nixosModules.holo-nats-server
```
steveej, line 55: not needed since the host-agent spawns NATS
```suggestion
```

PR #63, opened 2025-02-14
44 api for scheduling workloads

zeeshan595:
Closes [#44](https://github.com/Holo-Host/holo-host-private/issues/44)


rust/services/workload/src/lib.rs
```diff
@@ -165,7 +182,61 @@ impl WorkloadApi {
...
+        if assigned_host.remaining_capacity.cores < workload.system_specs.capacity.cores {
+            return false;
+        }
+
+        true
+    }
+
+    // looks through existing hosts to find possible hosts for a given workload
+    // returns the minimum number of hosts required for workload
+    pub async fn find_hosts_meeting_workload_criteria(
```
steveej, line 206: WDYT about a trait for this functionality so we can more easily mock and refactor it?

specifically i'm trying to avoid the tight coupling to mongodb for any of this functionality.

rust/services/workload/src/lib.rs
```diff
@@ -165,7 +182,61 @@ impl WorkloadApi {
...
+            },
+            doc! {
+                // the maximum number of hosts returned should be the minimum hosts required by workload
+                // sample randomized results and always return back atleast 1 result
+                "$sample": std::cmp::min(workload.min_hosts as i32, 1)
+            },
+        ];
+        let results = self.host_collection.aggregate(pipeline).await?;
+        if results.is_empty() {
+            return Err(anyhow!(
```
steveej, line 230:
```suggestion
            anyhow::bail!(
```

rust/util_libs/src/db/schemas.rs
```diff
@@ -25,52 +25,93 @@ pub use String as DeveloperJWT;
 pub use String as SemVer;
 
 // Providetype Alias for MongoDB ID (mongo's automated id)
-pub use String as MongoDbId;
+pub use bson::oid::ObjectId as MongoDbId;
```
steveej, line 28: is there anything wrong with using `ObjectId`? could this schema be reused for a different db type? if not, i'm interested in us finding an abstraction level that enables that

steveej (commented):
high level review to discuss the introduction of code that's tightly coupled to mongodb in other business domains. i highly suggest we decouple that logic from the data layer.

JettTech (commented):
Just adding a few additional questions and notes of things to handle either in this PR or in a follow-up.

rust/services/workload/src/lib.rs
```diff
@@ -83,12 +85,12 @@ impl WorkloadApi {
...
                     );
                     let updated_workload = schemas::Workload {
-                        _id: Some(workload_id),
+                        _id: Some(MongoDbId::from_str(&workload_id)?),
                         ..workload
                     };
                     Ok(types::ApiResult(
                         WorkloadStatus {
-                            id: updated_workload._id,
+                            id: updated_workload._id.map(|oid| oid.to_hex()),
```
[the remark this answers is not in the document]
zeeshan595: Well, when communicating with mongodb you should be using `bson::oid::ObjectId`. MongoDb's ID is not a string
In the actual json this is how it is stored
```
{ _id: { $oid: "super-cool-id" } }
```
Mongodb Also, treats the _id as it is an ObjectId. 
Doing queries and passing a String instead of ObjectId should panic / throw an error.

rust/util_libs/src/db/schemas.rs
zeeshan595: _id is always a `bson::oid::ObjectId`. Making `_id` a string will return no results if you try to use it to query it

image
image

If you want an abstraction I would say to use an ORM or a query builder. However, most of the popular ORMs / Query builders in rust do not support mongodb. So I would say just because we are using MongoDb we are kind of locked in with it.

https://www.sea-ql.org/SeaORM/docs/generate-entity/sea-orm-cli/#generating-entity-files
https://diesel.rs/guides/getting-started

rust/services/workload/src/lib.rs
zeeshan595: I think we are already tightly coupled with mongodb just because no ORM or query builder in rust supports MongoDb and we use their official rust package. Trying to abstract it and use it ourselves seems kind of silly to me. I would rather use sea-orm with postgres.

MongoDb and SQL like database have fundamentally different queries. MongoDb incentives using their aggregate function which is an array of json objects. While most other db's use a string query. You will have to re-write every query so it is accepted in a SQL like database.

rust/services/workload/src/lib.rs
```diff
@@ -110,21 +112,28 @@ impl WorkloadApi {
...
                         .await?;
+
                     log::info!(
                         "Successfully updated workload. MongodDB Workload ID={:?}",
                         workload._id
                     );
                     Ok(types::ApiResult(
                         WorkloadStatus {
-                            id: workload._id,
+                            id: workload._id.map(|oid| oid.to_hex()),
```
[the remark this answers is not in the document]
zeeshan595: I don't see a problem with that. However, it is common practice to return a string rather than a `bson::oid` object. So should keep a note of that.

rust/services/workload/src/lib.rs
zeeshan595: Let's move this discussion to another PR

rust/util_libs/src/db/schemas.rs
zeeshan595: Let's move this discussion to another PR

rust/services/workload/src/lib.rs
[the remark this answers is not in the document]
zeeshan595: Let's move this discussion to another PR

rust/services/workload/src/lib.rs
```diff
@@ -165,7 +182,61 @@ impl WorkloadApi {
         .await)
     }
 
+    // verifies if a host meets the workload criteria
+    pub fn verify_host_meets_workload_criteria(
```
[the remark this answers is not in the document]
zeeshan595: Moved to inventory, Remove this function from here

rust/util_libs/src/db/schemas.rs
steveej: clarifiation, i just meant to use `ObjectId` directly instead of giving it an alias.

zeeshan595: Yea, that makes sense. I can modify the PR to do that instead

rust/services/workload/src/lib.rs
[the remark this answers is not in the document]
zeeshan595: Moved to inventory, Remove this function from here

JettTech (approved):

PR #58, opened 2025-02-03
host-agent: multi-agent integration test and code fixes/improvements for it

steveej:
fixes https://github.com/Holo-Host/holo-host-private/issues/96

PR #64, opened 2025-02-17
Feature Proposal: reuse base client

JettTech:
Proposal for separating out and re-using the host agent client.
> NB: This is an extension of my feedback for pr #58 and is set to merge into that as the base branch


rust/clients/host_agent/src/workload_manager.rs
```diff
@@ -11,83 +11,45 @@
...
-    .await
-    .map_err(|e| anyhow::anyhow!("connecting to NATS via {nats_url}: {e}"))?;
+    let workload_service = host_client.add_js_service(workload_stream_service).await?;
 
     // ==================== DB Setup ====================
     // Create a new MongoDB Client and connect it to the cluster
     let mongo_uri = get_mongodb_url();
-    let client_options = ClientOptions::parse(mongo_uri).await?;
-    let client = MongoDBClient::with_options(client_options)?;
+    let db_client_options = ClientOptions::parse(mongo_uri).await?;
```
mattgeddes, line 44: Why is the host agent making calls to MongoDB? The only transport in/out of the agent should be over NATS and we're not running MongoDB on the holoports.

rust/clients/host_agent/src/main.rs
```diff
@@ -49,27 +50,28 @@ async fn main() -> Result<(), AgentCliError> {
...
 
-    let host_client = workload_manager::run(
+    let host_client = host_client::run(
         "host_id_placeholder>",
         &args.nats_leafnode_client_creds_path,
+        args.nats_connect_timeout_secs,
     )
     .await?;
 
+    workload_manager::run(host_client.clone()).await?;
```
steveej, line 69: what are the implications of cloning the client? is the client multi-threading safe?

rust/util_libs/src/nats_js_client.rs
```diff
@@ -236,11 +237,24 @@ impl JsClient {
...
+            &params.name,
+            &params.description,
+            &params.version,
+            &params.service_subject,
+        )
+        .await?;
+
         let mut current_services = self.js_services.unwrap_or_default();
-        current_services.extend(js_services);
+        current_services.push(new_service.clone());
```
steveej, line 254: what are the implications of cloning the service?

rust/util_libs/src/nats_js_client.rs
```diff
@@ -236,11 +237,24 @@ impl JsClient {
...
         self.js_services = Some(current_services);
-        self
+
+        Ok(new_service)
```
steveej, line 257: is there ever a use-case for needing this in the caller? the most important aspect here seems to me that the service is added to the current services.

steveej (commented):
i'm not clear on the implications of this and AFAIU we don't have any tests yet that cover data exchange for the host-agent itself, only for the underlying NATS connectivity.

the main point for me is to understand the implications of cloning a client. will it be a new, independent client that uses the configuration of the cloned one as a starting point?

rust/clients/host_agent/src/workload_manager.rs
JettTech: Yes, I agree.  This PR is not adding that functionality, but merely working with the baseline that is currently there.
(This baseline was a rough draft in support of testing from prior to the holidays and was tech debt.)

The code that removes this /handles the tech debt is in pr #53.  This code hasn't yet been merged into main, so new PRs still have this old tech debt in them.

I'm waiting on PR #63 to be merged into main prior to having this PR #53 reviewed in order to avoid having to handle a nasty merge conflict more than once between that  PR #53  and #63.

JettTech:
> AFAIU we don't have any tests yet that cover data exchange for the host-agent itself, only for the underlying NATS connectivity.

Actually, my goal for pr #62 is to include the testing of data exchange between the orchestrator and the host-agent.  What do you think about co-working on that this week and moving it to completion so that PRs like this can use that as a testing baseline?

rust/util_libs/src/nats_js_client.rs
JettTech: Yes.  This is what is referenced in the client file itself when setting up the stream consumers.  See an example here from this PR: https://github.com/Holo-Host/holo-host/blob/43dd00f4c316d4741c64d1d26b0c7280aaae691a/rust/clients/host_agent/src/workload_manager.rs#L53

rust/clients/host_agent/src/main.rs
JettTech: Yes, at its core, the client is the `async_nats::Client`, which thread-safe. Cloning this client does not create a new connection, but instead all cloned instances share the same underlying  TCP connection.  In terms of client connectitivy, the connection remains open as long as at least one clone of `Client` exists.  Once all clones are closed/dropped, the connection with the NATS server will close.

rust/util_libs/src/nats_js_client.rs
JettTech: The service is typed as `JsStreamService`.  Cloning the `JsStreamService` struct does not create an entirely new service, but instead shares the same underlying state since it uses Arc<RwLock<T>>` for its fields: `js_context` (the  jetstream client), `stream`, and `local_consumers`.  

Other things to note:
  - The `local_consumers` are configured as `PullConsumers`, which fetch messages manually and thereby prevent race conditions in message processing.
  - Modifications in one clone affect all other clones and there should be no risk of data inconsistency from cloning itself ..that said concurrent modifications need to ensure proper locking.
  
 Possible caveat:
   - If too many writes occur at the same time, they will be queued due to RwLock, potentially causing a slight delay.

JettTech:
> the main point for me is to understand the implications of cloning a client. will it be a new, independent client that uses the configuration of the cloned one as a starting point?

These are good questions, and I went to some detail to answer the outstanding Qs about the implications of cloning the nats client and the js stream service.   Let me know if you have any further Qs about this. :)

rust/util_libs/src/nats_js_client.rs
JettTech: Ok, just updated this to avoid the additional clone. :+1:

rust/util_libs/src/nats_js_client.rs
JettTech: As noted above, I removed the clone in this method.  On review, I do think it's still more preformant to just break up the adding and retrieval of the service into two different calls.

rust/clients/host_agent/src/main.rs
JettTech: One more note about the Nats `Client` and its clones to keep in mind.  Calling `drain` or `close` on any one of the instances/clones, closes the underlying connection on which all instances are based. This is because all clones are just references to the same connection.

steveej (commented):
> > AFAIU we don't have any tests yet that cover data exchange for the host-agent itself, only for the underlying NATS connectivity.
> 
> Actually, my goal for pr #62 is to include the testing of data exchange between the orchestrator and the host-agent. What do you think about co-working on that this week and moving it to completion so that PRs like this can use that as a testing baseline?

yes, let's do that. my preference for this PR is to postpone it until we have a test that verifies the functionality that's potentially affected by these changes.

PR #66, opened 2025-02-18
98 enforce issue linking

zeeshan595:
Closes [#98](https://github.com/holo-host/holo-host-private/issues/98)


steveej (commented):
this works for cases where an issue is justified. i'm making a case that there's exceptions to that in a [PR towards _this_ PR](https://github.com/Holo-Host/holo-host/pull/68).

you seem to agree with the existence of exceptions as _this_ PR itself uses a trick to satisfy the check :laughing:

zeeshan595:
> this works for cases where an issue is justified. i'm making a case that there's exceptions to that in a [PR towards _this_ PR](https://github.com/Holo-Host/holo-host/pull/68).
> 
> you seem to agree with the existence of exceptions as _this_ PR itself uses a trick to satisfy the check 😆

I don't believe this PR uses a trick to satisfy the check. But since we all can't agree on this. I can agree to add a skip label so you can use the skip label to skip the ci check :)

zeeshan595:
out of date

PR #74, opened 2025-02-21
refactor/util-libs

JettTech:
Contributes to #130


rust/clients/host_agent/src/workload_manager.rs
```diff
@@ -89,40 +94,40 @@ pub async fn run(
...
+            handler: EndpointType::Async(workload_api.call(
                 |api: WorkloadApi, msg: Arc<Message>| async move { api.start_workload(msg).await },
             )),
-            None,
-        )
+            response_subject_fn: None,
+        })
         .await?;
 
     workload_service
```
zeeshan595, line 107:
```rust
workload_service_add_consumer("install_workload", "start_workload", api.start_workload);
```
Should make a new function which calls `add_consumer`. There is a lot of code repetition

rust/util_libs/src/db/schemas.rs
```diff
@@ -46,24 +46,24 @@ pub struct User {
     pub metadata: Metadata,
     pub jurisdiction: String,
     pub permissions: Vec<UserPermission>,
-    pub user_info_id: Option<ObjectId>,
-    pub developer: Option<ObjectId>,
-    pub hoster: Option<ObjectId>,
+    pub user_info_id: Option<ObjectId>, // *INDEXED*
+    pub developer: Option<RoleInfo>,    // *INDEXED*
```
zeeshan595, line 50: Really happy we agreed to do this. This way we can have the `user` entry as a unique identity when communicated with any API

rust/util_libs/src/db/schemas.rs
```diff
@@ -197,39 +193,42 @@ pub enum WorkloadState {
...
 pub struct Workload {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub _id: Option<ObjectId>,
     pub metadata: Metadata,
-    pub state: WorkloadState,
-    pub assigned_developer: ObjectId, // *INDEXED*, Developer Mongodb ID
+    pub assigned_developer: ObjectId, // *INDEXED*
     pub version: SemVer,
     pub nix_pkg: String, // (Includes everthing needed to deploy workload - ie: binary & env pkg & deps, etc)
     pub min_hosts: u16,
```
zeeshan595, line 227: One thing I noticed is that in mongodb it does not have all types available in rust. So I would **avoid** using any `unsigned` field because in mongodb you cannot have `unsigned` field.

I would stick to using `int32` and `f32` for numbers when creating a schema for mongodb. And only using `i64` and `f64` when necessary.

Here are the conversions for mongodb

| Rust Type               | MongoDB Type          |
|-------------------------|----------------------|
| `i32`                  | `Int32`              |
| `i64`                  | `Int64`              |
| `u32`                  | `Int32`              |
| `u64`                  | `Int64` (if ≤ i64::MAX, otherwise potential loss) |
| `f32`                  | `Double` (stored as f64) |
| `f64`                  | `Double`             |
| `bool`                 | `Boolean`            |
| `String`               | `String`             |
| `&str`                 | `String`             |
| `Vec<u8>`              | `Binary`             |
| `Bson` (enum)          | Various BSON types   |
| `Option<T>`            | `Null` or `T`        |
| `chrono::DateTime<Utc>` | `DateTime`          |
| `ObjectId`             | `ObjectId`           |
| `Uuid`                 | `Binary (UUID subtype)` |
| `serde_json::Value`    | `Document` or other BSON type |
| `Vec<T>`               | `Array`              |
| `HashMap<String, T>`   | `Document`           |
| `BTreeMap<String, T>`  | `Document`           |

zeeshan595 (changes requested):

rust/util_libs/src/db/schemas.rs
JettTech: That's for researching this and calling it out. I'll update. :+1:

rust/clients/host_agent/src/workload_manager.rs
JettTech: Let's separate out this into a different PR.  I'm planning on making another PR to separate out tests too, so I'll prob just make another small PR with any other add'l refactor clean-ups.

zeeshan595: Ignore this because it is really annoying in rust and I am not sure if this is even possible because rust has limitations around reflection

zeeshan595 (approved):

PR #77, opened 2025-02-21
feature/orchestrator-workload-client

JettTech:
Contributes to [#43](https://github.com/Holo-Host/holo-host-private/issues/43)


mattgeddes (approved):

rust/clients/orchestrator/src/workloads.rs
```diff
@@ -0,0 +1,183 @@
...
+    // NB: These subjects are published by external Developer (via external api), the Nats-DB-Connector, or the Hosting Agent
+    let workload_service = orchestrator_workload_client
+        .get_js_service(WORKLOAD_SRV_NAME.to_string())
+        .await
+        .ok_or(anyhow!(
+            "Failed to locate Workload Service. Unable to spin up Orchestrator Workload Service."
+        ))?;
+
+    // Published by Developer
+    workload_service
```
zeeshan595, line 84:
```rust
workload_service_add_consumer("install_workload", "start_workload", api.start_workload);
```
Should make a new function which calls `add_consumer`. There is a lot of code repetition

zeeshan595 (changes requested):

zeeshan595 (approved):

rust/clients/orchestrator/src/workloads.rs
zeeshan595: Ignore this because it is really annoying in rust and I am not sure if this is even possible because rust has limitations around reflection

PR #80, opened 2025-02-24
132: Add version and git hash/status to agent CLI

mattgeddes:
Closes #issue-number


zeeshan595 (approved):

rust/clients/host_agent/build.rs
```diff
@@ -0,0 +1,20 @@
+use std::process::Command;
+
+fn main() {
+    let output = Command::new("git")
+        .args(&["rev-parse", "HEAD"])
```
steveej, line 5: `.git` is not part of the source directory in nix flakes so this will not work as intended. which explains the error on CI: https://buildbot-nix-0.infra.holochain.org/#/builders/416/builds/370

nix itself knows about the revision and there's a way to pass it through, maybe via an environment variable. if you rewrite this in a way that first tries to read the environment and only falls back to calling `git` we can make it work. i also suggest graceful handling of the inexistence of either.

rust/clients/host_agent/build.rs
```diff
@@ -0,0 +1,20 @@
...
+    let hash = String::from_utf8(output.stdout).unwrap();
+    println!("cargo:rustc-env=GIT_HASH={}", hash);
+
+    let status = Command::new("git")
+        .args(&["diff-index", "--quiet", "HEAD"])
+        .status()
+        .unwrap();
+    if status.success() {
+        println!("cargo:rustc-env=GIT_STATUS=clean");
+    } else {
```
steveej, line 17: this will consider a failed attempt at merely calling `git` to be in a dirty state. which is far from the truth for when `git` isn't even in PATH and `.git` is not present.

steveej: i suggest to differentiate between a failed execution of `git` and git reporting that the tree is dirty. for deployment in production via any of nix' tooling this is not relevant as per the discussion above.

rust/clients/host_agent/build.rs
mattgeddes: If you can tell me which environment variable(s) are present during the build, I can definitely change the above to use those instead.

rust/clients/host_agent/build.rs
mattgeddes: If it reports it as dirty and there's no hash in a controlled environment, it's pretty clear that it was a failure to run git. Let me know the environment variables and how to execute the build under that environment and I'll make some changes to the code.

rust/clients/host_agent/build.rs
steveej: i just stumbled upon this and am leaving it here for when we finish this RP:

```
      # the build process runs `git describe --long --always --dirty`
      fakeGit = writeShellScriptBin "git" ''
        echo "${finalAttrs.src.rev or finalAttrs.version}@nixpkgs"
      '';
```

rust/clients/host_agent/build.rs
steveej: the exit code of running `git status` in a directory that has no `.git` directory yields a `128` exit code:

```
$ git status
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
steveej@steveej-x13s ✗ (128) ~
```

we could forego having to interpret the exit code for this case by checking whether `.git` exists and not attempting to run `git` under a negative condition.

the question now is: what do we want to signal as a version/hash in that case? the easiest probably is the truth, i.e. `unknown`?

PR #81, opened 2025-02-24
63: Added some lightweight Nix state

mattgeddes:
Closes #63. Some of the heavier-weight Nix-related stuff we want to check on will be added to self-checks infrastructure.

### DIFF
--- a/rust/hpos-hal/src/inventory.rs
+++ b/rust/hpos-hal/src/inventory.rs
@@ -11,7 +11,7 @@ use procfs::{CpuInfo, FromBufRead};
 use serde_derive::{Deserialize, Serialize};
 use std::fmt::{self, Display};
 use std::io;
-use std::{fs, fs::File};
+use std::{fs, fs::read_to_string, fs::File};
 use thiserror::Error;
 use thiserror_context::{impl_context, Context};
 
@@ -88,6 +88,8 @@ pub struct HoloSystemInventory {
     pub kernel_version: String,
     /// OpenSSH Host public keys.
     pub ssh_host_keys: Vec<SSHPubKey>,
+    /// Small amounts of static state gathered from Nix
+    pub nix_state: Option<HoloNixState>,
 }
 
 /// A data structure representing an OpenSSH public key. When stored, each key is a single line of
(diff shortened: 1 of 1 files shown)


rust/hpos-hal/src/inventory.rs
```diff
@@ -105,6 +107,60 @@ pub struct SSHPubKey {
...
+    /// Canonicalisation of `/run/current-system` symlink
+    pub current_system: String,
+}
+
+impl HoloNixState {
+    const NIX_CURRENT_SYSTEM_FILE: &str = "/run/current-system";
+    // If we ever decide to enable `use-xdg-base-directories`, we'll need this to point to
+    // `$XDG_STATE_HOME/nix/channels` instead.
+    const NIX_CHANNELS_FILE: &str = "/root/.nix-channels";
+    fn from_host() -> Self {
```
zeeshan595, line 124: Could you add some tests for this?

zeeshan595 (changes requested):

rust/hpos-hal/src/inventory.rs
mattgeddes: Sure. Did you have anything specific in mind? I can put some wrapper functions around the constant path name, have the test(s) create a dummy channel file and use this code to parse it if we think there's value there.

steveej (commented):
i propose the following as additions to this PR as they're lightweight enough IMO:

* `/root/.nix-defexpr/channels/holo-nixpkgs/.git-version` as a blob
* `/var/lib/zerotier-one/identity.public` as a blob
* `/etc/nixos` as a compressed tarball. i expect this a few KB max. in my case it's 711B using `tar --dereference --hard-dereference -cavjf etcnixos.tbz2 /etc/nixos/*.nix`
* value of the NIX_PATH env var
* output of `systemctl status holo-nixpkgs-auto-upgrade.service` or equivalent
* output of `systemctl status auto-update.timer` or equivalent

having these in a follow-up is fine with me.
i just think without these the requirements discussed [in this issue](https://github.com/Holo-Host/holo-host-private/issues/63) aren't covered.

rust/hpos-hal/src/inventory.rs
zeeshan595: Just some unit tests would be fine by me. Just anything to run through a happy path and validate it works as intended

zeeshan595 (changes requested):

PR #84, opened 2025-02-27
131: restructure db and nats tests

JettTech:
Closes #131


JettTech:
@steveej - can you take a look at the CI and help determine what is causing the rust nats tests to continue to slow and never run?

FYI: Whenever I run them locally, they pass:
image

..., and in the CI I don't see any log output about what is delaying them.

steveej:
> @steveej - can you take a look at the latest CI build and help determine what is causing the rust nats tests to continue to slow and never run?

as a first pass let's prevent the tests can't run forever: https://github.com/Holo-Host/holo-host/pull/95

i haven't dug into the root cause of why it hangs. intuitively i''d point looking into concurrency leading to a port collision.

steveej (commented):
> @steveej - can you take a look at the latest CI build and help determine what is causing the rust nats tests to continue to slow and never run?

as a first pass let's prevent the tests can't run forever: https://github.com/Holo-Host/holo-host/pull/95

i haven't dug into the root cause of why it hangs. intuitively i''d point looking into concurrency leading to a port collision.

JettTech:
> i haven't dug into the root cause of why it hangs. intuitively i''d point looking into concurrency leading to a port collision.

I found out the root cause, it was due to a issue with the helper function I had created to check if the ports were available.  I was relying on making a `lsof` cli command to check the ports, but apparently `lsof` wasn't available in the CI env. :facepalm: Once I removed it, the tests passed. 

BTW, I had previously updated the coded to run on dynamic ports and to run serially.  So these issues are now resolved. :blush:

JettTech:
> as a first pass let's prevent the tests can't run forever: https://github.com/Holo-Host/holo-host/pull/95

Good idea. ;)  ...and that PR is now approved.  Thanks for tackling that chore so quickly! :slightly_smiling_face:

rust/util_libs/src/db/tests/mod.rs
```diff
@@ -0,0 +1,4 @@
+#[cfg(not(target_arch = "aarch64"))]
```
zeeshan595, line 1: I am not a fan of this unless there is a specific reason to add this?
Makes it so in the future tests can never be run on `aarch64`

rust/util_libs/src/nats/tests/gen_leaf_agents.rs
```diff
@@ -0,0 +1,210 @@
...
+    fs::create_dir_all(&nsc_sys_account_path).expect("Failed to create nsc creds dir");
+
+    let nsc_user_account_path = format!(
+        "{}/{}/{}/",
+        NSC_CREDS_PATH, OPERATOR_NAME, USER_ACCOUNT_NAME
+    );
+    fs::create_dir_all(&nsc_user_account_path).expect("Failed to create nsc creds dir");
+
+    // Create operator and sys account/user
+    Command::new("nsc")
```
zeeshan595, line 51: I think creating a helper function is properly better here. My rule of thumb is that if you have to write the same code more than twice then you should think about abstracting it.

```rust
fn execute_nsc_command(args: Vec<String>) {
  Command::new("nsc")
  .args(args)
  .output()
  .expect(format!("failed to execute command: {}", args))
}

execute_nsc_command(vec!["env", "-s", "TEST_AUTH_DIR"])
```

rust/util_libs/src/nats/tests/gen_leaf_agents.rs
```diff
@@ -0,0 +1,210 @@
...
+    output = Command::new("nsc")
+        .args(["describe", "account", "--name", USER_ACCOUNT_NAME, "--raw"])
+        .output()
+        .expect("Failed to execute nsc command");
+    if !output.status.success() {
+        log::debug!("Command failed with status: {}", output.status);
+        std::process::exit(1);
+    } else {
+        // Read the command output and filter out lines containing "-----"
+        let output_str = String::from_utf8_lossy(&output.stdout);
```
zeeshan595, line 196: This is also happening multiple times and perhaps could be in a function instead?

rust/util_libs/src/nats/tests/jetstream_service.rs
```diff
@@ -0,0 +1,249 @@
...
+    // Verify consumer was created
+    let info = service.get_consumer_stream_info("test_consumer").await?;
+    assert!(info.is_some());
+
+    let _ = server.shutdown().await;
+    Ok(())
+}
+
+#[tokio::test]
+#[serial]
```
zeeshan595, line 113: Is there a reason why the tests cannot run in parallel?

zeeshan595 (changes requested):

rust/util_libs/src/nats/tests/jetstream_service.rs
JettTech: Yes, the nats test are run in parallel because each one spawns a new nats server against which it runs its tests and then cleans it up.  To avoid heavy compute demand and avoid an increased probability for port conflicts, these are run in sequence.

rust/util_libs/src/nats/tests/gen_leaf_agents.rs
JettTech: Yes, I agree... that's why I put the `// TODO: Clean up this function to abstract away redundancy and improve readability` comment on line 15 of this file.  But since this is for test setup and we are on a tight timeline, I thought it'd be better to have the tests in and iterate over these lingering clean-up items in the provided team code debt management intervals, rather than not have any tests for our utils crate.

rust/util_libs/src/nats/tests/gen_leaf_agents.rs
JettTech: Yes, I agree... that's why I put the `// TODO: Clean up this function to abstract away redundancy and improve readability` comment on line 15 of this file.  :) 

But since this is for test setup and we are on a tight timeline, I think it'd be better to proceed with merging these tests in and then iterate over these lingering clean-up items in the provided team code debt management intervals, rather than not have any tests for our utils crate.

I'm going to leave this comment unresolved so I can reference it the next time I have space for this level clean-up.

rust/util_libs/src/nats/tests/gen_leaf_agents.rs
JettTech: See above.

rust/util_libs/src/db/tests/mod.rs
JettTech: This is only used on tests that rely on the mongodb runner and it needed because mongodb's socket handling can be inconsistent or require different configurations on aarch64 and the runner is not set up to account for all these nuances atm.

Please note that we are still testing these tests in our CI, as we do have both an aarch64 and x86_64-linux tests in our CI, so the functionality for the db is still being tested.

JettTech:
Overtaken by pr #110

PR #89, opened 2025-03-03
treefmt: add TOML formatter, format repo

steveej:
note that we don't have a formatter check on CI for this, which.

(there's an upstream ticket for that: https://github.com/numtide/treefmt-nix/issues/317)


JettTech (approved):

JettTech:
Thanks for doing this!!

PR #103, opened 2025-03-12
add inventory machine_id as host identifier

JettTech:
Updates the host id pattern, setting to the host identifier to the inventory `machine_id`.  

This handles the passing of the id into the inventory and workload subject strings.  It also manages the upsert of the host record into the host collection during an inventory update, whenever the host does not yet exist.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -21,23 +21,23 @@ dependencies = [
 
 [[package]]
 name = "actix-http"
-version = "3.9.0"
+version = "3.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d48f96fc3003717aeb9856ca3d02a8c7de502667ad76eeacd830b48d2e91fac4"
+checksum = "0fa882656b67966045e4152c634051e70346939fced7117d5f0b52146a7c74c9"
 dependencies = [
  "actix-codec",
  "actix-rt",
  "actix-service",
  "actix-utils",
- "ahash",
  "base64 0.22.1",
  "bitflags 2.9.0",
  "brotli",
  "bytes",
  "bytestring",
- "derive_more 0.99.19",
+ "derive_more 2.0.1",
  "encoding_rs",
  "flate2",
+ "foldhash",
  "futures-core",
  "h2 0.3.26",
  "http 0.2.12",
@@ -49,7 +49,7 @@ dependencies = [
  "mime",
  "percent-encoding",
  "pin-project-lite",
- "rand 0.8.5",
+ "rand 0.9.0",
  "sha1",
  "smallvec",
  "tokio",
@@ -65,7 +65,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e01ed3140b2f8d422c68afa1ed2e85d996ea619c988ac834d255db32138655cb"
 dependencies = [
  "quote",
- "syn 2.0.99",
+ "syn 2.0.100",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


JettTech:
@steveej  - it looks like CI is now failing due to the same openssl issue I encountered locally.
https://buildbot-nix-0.infra.holochain.org/#/builders/416/builds/529

image

steveej:
thanks for this!

> it looks like CI is now failing due to the same openssl issue I encountered locally.

ah nice, i'm glad it shows up on CI, taking a look :monocle_face:

Cargo.lock
```diff
@@ -4104,15 +4154,6 @@ version = "0.1.6"
...
 checksum = "d05e27ee213611ffe7d6348b942e8f942b37114c00cc03cec254295a4a17852e"
 
-[[package]]
-name = "openssl-src"
-version = "300.4.2+3.4.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "168ce4e058f975fe43e89d9ccf78ca668601887ae736090aacc23ae353c298e2"
-dependencies = [
- "cc",
-]
```
steveej, line 4114: this is causing the build problems as it introduces the dependency on the build environment's openssl.

i'm going to revert the Cargo.lock changes for this PR and we can deal with updating it and adjusting the buildinputs separately.

rust/util_libs/db/src/schemas.rs
```diff
@@ -155,27 +159,52 @@ pub struct Host {
...
+    0
+}
+
+fn default_avg_latency() -> i64 {
+    0
+}
+
+fn empty_object_id() -> ObjectId {
+    ObjectId::from_bytes([0; 12])
+}
```
steveej, line 195: my order preference for this would be
* `impl Default for ...`
* inline with the struct
* defined as `const` items (can be `const fn` too)

rust/util_libs/db/src/schemas.rs
```diff
@@ -35,9 +35,13 @@ pub enum UserPermission {
 
 #[derive(Serialize, Deserialize, Clone, Debug, Default)]
```
steveej, line 36:
```suggestion
#[derive(Serialize, Deserialize, Clone, Debug, Default)]
#[serde(default)] 
```

i believe putting this on the container will be enough to replace all the field occurrences of this, except for the ones that don't have an `impl Default`.

rust/services/workload/src/types.rs
```diff
@@ -6,8 +6,10 @@ use std::collections::HashMap;
 use strum_macros::AsRefStr;
 
 #[derive(Serialize, Deserialize, Clone, Debug)]
-pub struct ObjectIdJSON {
+pub struct HostIdJSON {
     pub _id: ObjectId,
+    #[serde(skip_serializing_if = "String::is_empty")]
+    pub device_id: String,
```
steveej, line 12: i'd like us to commit to one semantic here:

steveej: 1. optional

```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub maybe_device_id: Option<String>,
```

steveej: 2. required

1. optional

```suggestion
    pub device_id: String,
```

steveej (commented):
thanks for this!

i've got a few comments.

and i think this could actually go directly to `main`!

rust/clients/orchestrator/src/admin_client.rs
```diff
@@ -14,9 +13,8 @@ pub async fn run(admin_creds_path: &Option<PathBuf>, nats_url: String) -> anyhow
 
     let credentials = admin_creds_path
         .to_owned()
-        .map(|creds| vec![Credentials::Path(creds)]);
-
-    // .ok_or(anyhow!("Failed to locate admin credential path."))?;
+        .map(|p| vec![Credentials::Path(p)])
+        .filter(|c| !c.is_empty());
```
steveej, line 17:
```suggestion
        .map(|p| vec![Credentials::Path(p)]);
```

since `Option::map` is only effective on `Some` values

JettTech:
Was forked and merged in with pr #105.

rust/services/workload/src/types.rs
JettTech: Let's go with 2.

PR #106, opened 2025-03-13
Main hot fixes

JettTech:
### Updates:
 - Fixes multi-service runtime issue 
 - Fixes inventory handling when device class file is not found
 - Adds  inventory amendment diffs


steveej (approved):
thanks!

i've added some more commits to address the issues that were caught by CI.

PR #108, opened 2025-03-17
README,nix(devshells/default): add `just` and introduce it

steveej:


mattgeddes (approved):
Thanks! This kind of thing looks to be really helpful and I can't wait to kick the tires on it!

steveej:
> This kind of thing looks to be really helpful and I can't wait to kick the tires on it!

agreed! i'm looking forward to getting feedback about the existing recipes. i just noticed that the links weren't present, that's fixed now. the more sophisticated commands currently live in #100, e.g. the `just dev-cycle` one cleans up and spins up the three main components of the platform in separate nixos-containers. that should work on any machine that has a working installation of `nix` and on a systemd-based linux distribution.

PR #118, opened 2025-03-21
feat: improved CLI/ENV args parsing, especially for NATS

steveej:
extracted from #100

### DIFF
--- a/rust/util_libs/nats/src/jetstream_client.rs
+++ b/rust/util_libs/nats/src/jetstream_client.rs
@@ -1,16 +1,21 @@
+use crate::types::NatsRemoteArgs;
+
 use super::{
     jetstream_service::JsStreamService,
-    leaf_server::LEAF_SERVER_DEFAULT_LISTEN_PORT,
     types::{
         Credentials, ErrClientDisconnected, EventHandler, EventListener, JsClientBuilder,
         JsServiceBuilder, PublishInfo,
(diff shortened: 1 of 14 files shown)


rust/util_libs/nats/src/jetstream_client.rs
```diff
@@ -27,24 +32,141 @@ impl std::fmt::Debug for JsClient {
...
+            })
             .ping_interval(p.ping_interval.unwrap_or(Duration::from_secs(120)))
-            .request_timeout(Some(p.request_timeout.unwrap_or(Duration::from_secs(10))))
+            .request_timeout(Some(p.request_timeout.unwrap_or(Duration::from_secs(1))))
             .custom_inbox_prefix(&p.inbox_prefix);
-        // .require_tls(true)
+
+        if let Some((user, pass)) = p.nats_remote_args.maybe_user_password()? {
+            connect_options = connect_options.user_and_password(user, pass);
+        }
```
JettTech, line 169: Can we unify this approach with the `p.credentials` below?  

Maybe we can just remove the credentials arg altogether and just add the token and nsc credentials file path options to the `NatsRemoteArgs`?

JettTech (approved):
Awesome updates! Thanks for the work!! :muscle:

I just left one small change request, but otherwise think this is ready to merge.

PR #119, opened 2025-03-22
feat(ham): custom addr, download as Bytes, cli/lib fn for finding installed app

steveej:
this is a codified learning in preparation for working making app interface calls.

extracted from #100

### DIFF
--- a/rust/ham/src/lib.rs
+++ b/rust/ham/src/lib.rs
@@ -5,16 +5,30 @@
 
 use anyhow::{Context, Result};
 use derive_builder::Builder;
-use holochain_client::AdminWebsocket;
-use holochain_conductor_api::AppInfo;
+use holochain_client::{
+    AdminWebsocket, AgentPubKey, AppWebsocket, AuthorizeSigningCredentialsPayload, CellInfo,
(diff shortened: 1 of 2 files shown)


rust/ham/src/lib.rs
```diff
@@ -103,52 +100,31 @@ impl Ham {
...
-            let bundle = holochain_types::app::AppBundle::decode(happ_bytes)
-                .context("decoding happ_bytes into an AppBundle".to_string())?;
-
-            #[allow(deprecated)]
-            let source = AppBundleSource::Bundle(bundle);
+            let source = AppBundleSource::Bytes(happ_bytes.to_vec());
             InstallAppPayload {
                 agent_key: Some(agent_key.clone()),
-                installed_app_id: None,
+                installed_app_id: maybe_installed_app_id,
```
JettTech, line 124: This may be something we wish to punt for later as this won't currently be an issue for us, but I do think that we should be passing in an `installed_app_id` bc if we don't pass it in, it'll be derived from the app name that's provided in the app manifest -- which we can't assume will be unique.

JettTech (approved):
I think this is good iterative start and knowing that we are already considering some different patterns that will restructure this flow while integrating the hc gateway work, I think this is good to merge. :)

PR #122, opened 2025-03-22
fix(hpos-agent/inventory): startup behavior and resiliency

steveej:
startup behavior: starting at `now` it would previously wait an entire interval before sending the inventory; solely based on the time check.

resiliency: never stop trying to send inventory messages. back off the time between attempts and reset the retry on every successful event.

---

i discovered the issues while working on #100.

alternative and preferred alternative to #120. therefore:

closes #120.

### DIFF
--- a/rust/clients/host_agent/src/hostd/inventory.rs
+++ b/rust/clients/host_agent/src/hostd/inventory.rs
@@ -9,7 +9,9 @@
   This client does not subject to or consume any inventory subjects.
 */
 
-use anyhow::Result;
+use std::{path::Path, time::Duration};
+
+use futures::TryFutureExt;
 use hpos_hal::inventory::HoloInventory;
 use inventory::INVENTORY_UPDATE_SUBJECT;
 use nats_utils::{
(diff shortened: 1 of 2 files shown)


rust/clients/host_agent/src/hostd/inventory.rs
```diff
@@ -26,75 +28,106 @@ pub fn should_check_inventory(
...
-    starting_inventory
-        .save_to_file(inventory_file_path)
-        .map_err(|e| {
-            ServiceError::internal(
-                e.to_string(),
-                Some("Failed to save host inventory to file.".to_string()),
-            )
-        })?;
-
     let one_hour_interval = tokio::time::Duration::from_secs(host_inventory_check_interval_sec);
```
JettTech, line 41:
```suggestion
    let interval = tokio::time::Duration::from_secs(host_inventory_check_interval_sec);
```
I do prefer the generic naming applied in the pr 100/120.

rust/clients/host_agent/src/hostd/inventory.rs
```diff
@@ -26,75 +28,106 @@ pub fn should_check_inventory(
...
     let check_interval_duration = chrono::TimeDelta::seconds(one_hour_interval.as_secs() as i64);
-    let mut last_check_time = chrono::Utc::now();
+    let mut last_check_time = chrono::Utc::now() - one_hour_interval;
```
JettTech, line 43:
```suggestion
    let mut last_check_time = chrono::Utc::now() - interval;
```

JettTech (approved):
LGTM :)

PR #125, opened 2025-03-23
feat(util_libs): tolerate anyhow errors, host/workload schema schemas

steveej:
* the macro now tolerates results that declare anyhow::Error for NATS
  consumer functions
* host and workload schema changes

extracted from #100

### DIFF
--- a/rust/util_libs/db/src/schemas.rs
+++ b/rust/util_libs/db/src/schemas.rs
@@ -13,14 +13,14 @@ use crate::mongodb::MutMetadata;
 /// use mongodb::Client;
 ///
 /// // Work with collections using the defined schemas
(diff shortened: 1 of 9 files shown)


rust/util_libs/db/src/schemas.rs
```diff
@@ -254,8 +254,6 @@ pub struct Host {
     pub metadata: Metadata,
     /// Unique identifier for the device
     pub device_id: String,
-    /// IP address of the host
-    pub ip_address: String,
```
steveej, line 258: @JettTech i'm not sure whether this removal is the most recent intention. it's reintroduced in #100 just in case we merge this one (#125) as is.

JettTech (approved):

PR #129, opened 2025-03-26
add nats to holo-gateway

JettTech:
Closes #144


JettTech:
This work has been moved into pr #132

PR #131, opened 2025-03-27
DNS Deployment Pipeline

zeeshan595:
Closes [142](https://github.com/Holo-Host/holo-host-private/issues/142)


rust/holo-dns/dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+WORKDIR /usr/src
+COPY . .
+
+RUN --mount=type=cache,target=/root/.cargo \
+    --mount=type=cache,target=/usr/src/target \
+    cargo build --release && \
+    cp /usr/src/target/release/holo-dns /usr/src/build
+
+FROM gcr.io/distroless/cc
```
mattgeddes, line 11: We should raise a bug to fix this later, but we shouldn't even need glibc. For the DNS service, the public API service and HTTP/NATS gateway, we should build them to the x86_64-unknown-linux-musl target. That'll link them statically, remove dependencies on glibc and drastically shrink the container size. It's something that we need to gate the release on though. So if we raise a bug for that, that'll suffice for now.

mattgeddes (approved):

PR #132, opened 2025-03-28
[fork of] 144 add nats to holo gateway steveej

steveej:
this is branched off #129 and rebased onto main.

this PR is for integrating the holo-gateway with the extra-container-dev environment, which also implies a nix package and nixos module.


rust/services/workload/src/host_api.rs
```diff
@@ -318,7 +318,15 @@ impl HostWorkloadApi {
                     let extra_container_path =
                         provision_extra_container_closure_path(&workload_path_toplevel.into())?;
 
-                    bash(&format!("extra-container destroy {extra_container_path}")).await?;
+                    let extra_container_subcmd = match desired_state {
+                        WorkloadState::Deleted => "stop",
+                        _ => "destroy",
```
steveej, line 323: reverse

```suggestion
                        WorkloadState::Deleted => "destroy",
                        _ => "stop",
```

mattgeddes (approved):
This looks good to me. Especially breaking some of the initial prototype code out into separate modules. Had I known things had changed in this way, I'd have created this commit (https://github.com/Holo-Host/holo-host/commit/b6025a02a8fae69923704c738a734833ef77c188) against this branch instead to make it easier to merge in....

rust/holo-gateway/src/nats_client.rs
```diff
@@ -0,0 +1 @@
+use crate::types;
```
JettTech, line 1: Do we still need this file? Can we remove it?

rust/holo-gateway/Cargo.toml
```diff
@@ -0,0 +1,49 @@
...
+
+[dev-dependencies]
+mock_utils = { workspace = true }
+http-body-util = "0.1.2"
+reqwest = { version = "0.12", features = ["json"] }
+
+[profile.release]
+lto = true
+strip = true
+codegen-units = 1
```
JettTech, line 49:
```suggestion
codegen-units = 1

[features]
broken = []
```

...to help with clippy

JettTech (approved):
Approved - Thanks for all the clean up and merging of this code with the code to support the host agent side.

rust/holo-gateway/Cargo.toml
steveej: my intention for that has been to resolve this differently. either by removing these tests and agreeing on a future direction for a substitute, or by agreeing on a modification of them for this PR.

PR #133, opened 2025-03-31
README: show some examples of building CI tests locally

steveej:


mattgeddes (approved):
Thanks for this! It's much appreciated! I'm running it now -- the syntax seems to work and I'll see what the results are. Thanks again.

PR #141, opened 2025-04-11
104 holo public api updated

zeeshan595:
pipeline

- [x] New Feature
- [ ] Bug Fix

### Description
- Initial setup for holo public api
- It contians auth and logging middleware with a health check endpoint
- Uses db utils
- Added mongodb & redis as background tasks inside github actions


zeeshan595:
Just a heads up. The tests are failing in build bot because the integration tests require a local instance of mongodb and redis to be running. Github actions runs those before running the tests but  build bot does not which is why the tests fail on build bot

mattgeddes (approved):

PR #145, opened 2025-04-17
193: Minor schema updates and added versioning

mattgeddes:
pipeline

- [ ] New Feature
- [ ] Bug Fix

### Description
Made some small schema updates to the DNS service and added versioning to support multiple versions


JettTech (approved):
LGTM!  One idea for future automation would be to extract out the schema version into a config file or the like to decouple it a bit from the code and be able to reuse the same binary across different deployments with only config differences.

PR #146, opened 2025-04-24
schema updated

zeeshan595:
pipeline

- [ ] New Feature
- [x] Bug Fix

### Description
- This PR improves the db schema to remove unnecessary fields that are not required and would duplicate data. It follows the rule of 1 source of truth and thus any field that gives duplicate data is removed.
- User's pubkey + pubkey_role has been made into a vec array
- User's now have more granular permmissions
- Users now have roles
- workload and hosts now refer to the user who created / owns them as `owner`


rust/util_libs/db/src/schemas/user_info.rs
```diff
@@ -25,6 +25,8 @@ pub struct UserInfo {
     pub given_names: String,
     /// User's family name
     pub family_name: String,
+    // User's jurisdiction
+    pub jurisdiction: String,
```
mattgeddes, line 29: Should this be something more complex than a String? Otherwise it's too difficult to ensure consistency and correctness. I was thinking perhaps an enum of all jurisdictions and then various groupings on top of that for larger jurisdictions just off the top of my head. All wrapped in a type with validation and easy test methods.

Not saying we should enumerate all of those jurisdictions ourselves. Is there anything out there that already exists? Perhaps we can even start with one of the existing internationalisation constructs and related crates if there isn't something close enough to what we want. I'll take a look too. Something that complements or is complemented by GeoIP will help in the future too, I think.

Also, should we call the field itself `geographic_jurisdiction` just in case we end up with some other jurisdiction constructs/policies that aren't related to geography? (Departments for example)

zeeshan595: I agree with that. I remember we did have something like that in the Old system so I would be in favour of that!

zeeshan595:
This is the query I ran on `registrations` which has `2k` records and `holoalpharegistrations` has `1k` records
```json
[
  {
    $match: { _id: ObjectId('6579ec32a85de7fbb288a821') }
  },
  {
    $lookup: {
      from: "holoalpharegistrations",
      localField: "oldHoloportIds._id",
      foreignField: "holoportIds.id",
      as: "alpha"
    }
  },
  {
    $unwind: "$alpha"
  }
]
````
image

I also ran the same query using `registrations` + `uptime_daily` with 1.5M records and the results are attached in the screenshot
```json
[
  {
    $match: {
      _id: ObjectId("6579ec32a85de7fbb288a821")
    }
  },
  {
    $lookup: {
      from: "uptime_daily",
      localField: "oldHoloportIds._id",
      foreignField: "holoportId",
      as: "uptime"
    }
  },
  {
    $unwind: "$uptime"
  }
]
```
image

rust/util_libs/db/src/schemas/user_info.rs
zeeshan595: This is done as part of the holo public api

zeeshan595:
I am going to close this PR and create a github issue to do this after humm happ changes have been done

PR #150, opened 2025-05-07
104 holo public api sync

zeeshan595:
pipeline

- [x] New Feature
- [ ] Bug Fix

### Description
This PR aims to sync the deployed version of the holo public api to the main branch.
It includes the following new endpoints

- login with apikey
- refresh auth token

- create api key
- get one api key
- get many api keys
- update api key
- delete api key

- upload blob

- create workload
- get workload
- delete workload

- create user (for internal user, registering using legacy system)

Note: tests are missing because there was a overhead because of the delays that caused the api to not be merged in and additional work to resolve conflicts from the branches. In addition, including tests would cause the already big PR massive. Therefore, the tests will be done as part of smaller PRs in the future

Since workloads use the id of a developer not a user currently. I added a temporary get call to get the developer. Once we have an `owner` in the database this call will not be needed.


rust/holo-public-api/src/controllers/blob/upload.rs
```diff
@@ -0,0 +1,205 @@
...
+            owner: owner.clone(),
+        }],
+    ) {
+        return HttpResponse::Forbidden().json(ErrorResponse {
+            message: "forbidden".to_string(),
+        });
+    }
+    let temp_file_identifier = bson::uuid::Uuid::new().to_string();
+    let mut hasher = blake3::Hasher::new();
+    let mut blob = match fs::File::create(format!("/tmp/{}", temp_file_identifier)) {
```
mattgeddes, line 78: Can we make this runtime configurable, and default to somewhere that isn't /tmp? On most Linux distributions /tmp is tmpfs (which is memory-backed, and therefore volatile).

rust/holo-public-api/src/controllers/blob/upload.rs
```diff
@@ -0,0 +1,205 @@
...
+                tracing::error!("{:?}", e);
+                return HttpResponse::InternalServerError().json(ErrorResponse {
+                    message: "Internal server error".to_string(),
+                });
+            }
+        }
+    };
+
+    // copy blob to the correct location
+    match fs::copy(
```
mattgeddes, line 158: A better pattern than write->copy->delete might be to write->rename. This will mean writing to a temporary file on the same filesystem and then renaming to the destination filename. The copy you're doing here could potentially be expensive, as you're writing the file twice and it's not atomic. 

For example, the current pattern would look like the equivalent of this:
 - write to /tmp/xxx
 - copy to /srv/blobstore/yyy
 - remove /tmp/xxx

In this case, you write to /tmp and then ready from /tmp while also writing again to /srv. For a large file, this could be a lot of extra unnecessary I/O. Also, while you're writing to /srv/blobstore/yyy, if another process attempts to write to the same file, the results (according to POSIX) are undefined.

The pattern I would suggest would look more like this:
 - write to /srv/blobstore/.yyy.tmp
 - rename to /srv/blobstore/xxx

In this case, you're only writing the file data once to a unique filename, thus avoiding extra I/O and the issue of parallel processes writing to the same file. The rename is atomic, so if two processes try to do the same thing, you still end up with consistent results.

rust/holo-public-api/src/controllers/blob/upload.rs
```diff
@@ -0,0 +1,205 @@
...
+        Err(e) => {
+            tracing::error!("{:?}", e);
+            return HttpResponse::InternalServerError().json(ErrorResponse {
+                message: "Internal server error".to_string(),
+            });
+        }
+    };
+
+    // create metadata blob
+    let mut metadata_file = match fs::File::create(format!("{}/{}.json", file_location, hash_hex)) {
```
mattgeddes, line 184: Just to be safe, it's probably worth following the same pattern as above, where we write to a temporary file and then rename. That said, we're likely to only be writing a single block of I/O for the metadata, so the odds of some of the problems mentioned above are less likely.

rust/holo-public-api/src/controllers/blob/upload.rs
mattgeddes: I hope I've explained the above clearly enough. If not, we can certainly run through it for a few minutes on one of the scheduled Monday morning calls.

rust/holo-public-api/src/controllers/user/create_user.rs
```diff
@@ -0,0 +1,228 @@
...
+    public_key: String,
+
+    /// the role of the user
+    role: PublicKeyRoleInfo,
+}
+
+#[derive(Serialize, Deserialize, ToSchema)]
+pub struct UserInfo {
+    /// the email of the user, this can be used as a login flow
+    #[schema(example = "john.doe@email.com")]
```
mattgeddes, line 34: A little bit nit-picky perhaps, but RFC 6761 reserves the domains `example.com`, `example.net` and `example.org`  domains for use in documentation and examples, not `email.com`...

rust/holo-public-api/src/controllers/workload/create_workload.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub id: String,
+}
+
+#[derive(OpenApi)]
+#[openapi(paths(create_workload))]
+pub struct OpenApiSpec;
+
+#[derive(Serialize, Deserialize, Debug, Clone, ToSchema)]
+#[serde(rename_all = "snake_case")]
+pub struct CreateWorkloadRequest {
```
mattgeddes, line 24: This definition is fine for now, but is almost certainly going to change drastically once we get into defining workload policies. We won't be exposing details around specific system resources or specific assigned hosts (much in the same way that AWS doesn't).

Is there a way that we can mark this as being hidden or private or something for now? I'm hoping that we don't have any customers come across this API and start trying to use it before we're ready. What are our options for hiding/gating APIs like this while we're still working on them?

rust/holo-public-api/src/providers/docs.rs
```diff
@@ -0,0 +1,68 @@
...
+    docs.info.version = "0.5.3".to_string();
+    docs.info.description = Some(
+        "Holo Public API is a RESTful API that allows you to interact with the Holo platform.
+You can manage your API keys, workloads, blobs, user accounts, and more.
+"
+        .to_string(),
+    );
+    // auth tag
+    let mut auth_tag = utoipa::openapi::Tag::new("Auth");
+    auth_tag.description = Some(
```
mattgeddes, line 23: Thanks for being thorough with the docs.

rust/util_libs/db/src/schemas/jurisdiction.rs
```diff
@@ -0,0 +1,495 @@
...
+    Debug,
+    Clone,
+    PartialEq,
+    Eq,
+    Hash,
+    Default,
+    ToSchema,
+)]
+#[strum(serialize_all = "title_case")]
+pub enum Jurisdiction {
```
mattgeddes, line 19: Thanks for making this more complex than a string. I'm still wondering whether there's something already out there in an existing crate, and also that covered different levels of jurisdiction. For example, we have country names here, which makes sense -- Germany and Austria are both jurisdictions, but the European Union is also a jurisdiction that we could have to deal with in workload and other policies. 

I don't think that we ought to try to implement all of that complexity -- I'd just be surprised if there wasn't already a crate out there that handled this, or that we could extend. I'll take a look too and see what I can find.

rust/holo-public-api/src/controllers/workload/create_workload.rs
zeeshan595: Yep, I can create a private and a public docs

rust/util_libs/db/src/schemas/jurisdiction.rs
zeeshan595: I agree with this and I will look into another API. I just copy pasted the list we had in the legacy system. Thought it would be good for now. Also, makes the migration easy

rust/holo-public-api/src/controllers/workload/create_workload.rs
zeeshan595: This is done, now there are internal docs which we can disable for public endpoints

rust/holo-public-api/src/controllers/user/create_user.rs
zeeshan595: Done

mattgeddes (approved):
The latest changes seem fine to me

PR #172, opened 2025-05-15
116 add rate limiting

zeeshan595:
pipeline

- [x] New Feature
- [ ] Bug Fix

- adds global rate limit for all endpoints
- integration test for global rate limit
- function for adding custom rate limits per endpoint
- rate limit is done by ip address, access token or the developer can provide their own key for custom rate limit on endpoint
- improved environment variables config, added default values for better developer experience

[116](https://github.com/orgs/Holo-Host/projects/10?pane=issue&itemId=97759210&issue=Holo-Host%7Cholo-host-private%7C116)


rust/holo-public-api/src/middlewares/limiter.rs
```diff
@@ -0,0 +1,91 @@
+use actix_web::{
+    body::{BoxBody, MessageBody},
+    dev::{ServiceRequest, ServiceResponse},
+    http::StatusCode,
+    middleware::Next,
+    web, Error,
+};
+use deadpool_redis::{redis::AsyncCommands, Pool};
```
mattgeddes, line 8: Is the redis instance shared across all of the API service instances? How/where is it being deployed?

mattgeddes (approved):

rust/holo-public-api/src/middlewares/limiter.rs
zeeshan595: Yep, it is a central digital ocean instance basically the same way mongodb is hosted. Ideally we want to deploy it via pulumi but the whole api right now does not have any infra. I manually created it via digital ocean portal

Both mongodb and redis are connected via a private network, they also have a whitelist rule so only the api (+allowed services) are allowed to access them

PR #175, opened 2025-05-17
116 rate limiting bug

zeeshan595:
pipeline

- [ ] New Feature
- [x] Bug Fix

### Description
- Nginx proxy can change the ip address header causing all of the requests to be coming from the same address
- TTL was not being set properly for the keys

### DIFF
--- a/rust/holo-public-api/src/middlewares/limiter.rs
+++ b/rust/holo-public-api/src/middlewares/limiter.rs
@@ -86,9 +86,11 @@ pub async fn rate_limiter_middleware(
             );
         }
 
-        conn.set_ex(key, count + 1, window as u64)
-            .await
-            .unwrap_or(());
+        if count == 0 {
+            conn.set_ex(key, 1, window as u64).await.unwrap_or(());
+        } else {
+            conn.incr(key, count + 1).await.unwrap_or(());
+        }
     }
     let resp = next.call(req).await?;
     Ok(resp.map_into_boxed_body())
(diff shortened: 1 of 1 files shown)


rust/holo-public-api/src/middlewares/limiter.rs
```diff
@@ -14,10 +14,18 @@ pub async fn rate_limiter_middleware(
...
     next: Next<impl MessageBody + 'static>,
 ) -> Result<ServiceResponse<BoxBody>, Error> {
+    // get ip from x-forward-for header or peer address
+    // nginx uses x-forwarded-for header to pass the real ip
     let ip = req
-        .peer_addr()
-        .map(|addr| addr.ip().to_string())
-        .unwrap_or_default();
+        .headers()
+        .get("X-Forwarded-For")
```
JettTech, line 21: This looks good, just a reminder to do a quick security check to ensure this is no possibility for someone to spoof this forwarded addr.

JettTech (approved):
One comment, otherwise lgtm!

rust/holo-public-api/src/middlewares/limiter.rs
zeeshan595: Okay, I did some research on this and we should be splitting by `,` and taking the last item. When the request comes in and is forwarded. Anything inside the `X-Forwarded-For` header is prepended. So if I pass 192.0.0.1 in the header and send the request it will end up being `192.0.0.1, 104.248.91.77`  where `104.248.91.77` would be the actual ip which is equal peer_addr

PR #178, opened 2025-06-03
Features: Db streaming + HC Versioning + Container Private Networks

JettTech:
- [x] New Feature(s)
- [ ] Bug Fix

### Description
- adds hc versioning per-container (based on explicit supported versions config in module definition)
- adds private networking per-container (using socat for port forwarding)
- adds live db streaming for the workload collection
- restructures & renames orchestrator workload service methods to  no longer be oriented
around nats<>db connector 
- removes smart host selection based on capacity 

> closes https://github.com/Holo-Host/holo-host-private/issues/217


nix/packages/extra-container-dev.nix
```diff
@@ -1,18 +1,14 @@
...
 {
   inputs,
   system,
   flake,
   perSystem,
   nixpkgs ? inputs.nixpkgs-2411,
-}:
-
-let
+}: let
```
mattgeddes, line 8: Was this change intended? It just seems like a whitespace change.

mattgeddes (commented):
I've read through the whole review and it all looks fine syntactically, but I can't figure out which issue this is resolving. I thought it was one thing, but then there's also a ton of formatting changes (especially to Nix stuff unrelated to the database location stuff I thought it was for). But it may also be addressing other things like using a streaming call for the 1MB request size issue or maybe not? I'm having trouble keeping all of the context in my head as I read through it and applying it to the issue it's supposed to be resolving. Can you give me some context for what we're fixing?

nix/packages/extra-container-dev.nix
JettTech: FYI - This base pr has now be merged into main and the base branch had been adjusted to point to main again automatically.

zeeshan595:
pipeline is failing

mattgeddes (approved):

mattgeddes (approved):

mattgeddes (approved):

PR #183, opened 2025-07-11
Bug 254: Added memory and OS release info to inventory

mattgeddes:
This resolves #254 in holo-host-private. I've added some initial information about memory modules present. It could use some additional work later, but works fine on my holoport and some other hosts I have around the house. I've also included the output of `nix --version` and `nixos-version`.


JettTech (approved):
lgtm :ship:

PR #186, opened 2025-07-25
Hot fix workload deployment

JettTech:
pipeline

- [x] New Feature
- [x] Bug Fix

### Description
 - fix: adds backward compatibility for the `happ_binary` field in the `workload` schema when using a json/web-client.
 - feature: adds in the nix configuration to support a default hc file


mattgeddes (dismissed):
Thanks Lisa!

PR #195, opened 2025-08-15
Feat/schema holo public api

zeeshan595:
pipeline

- [ ] New Feature
- [ ] Bug Fix

### DIFF
--- a/rust/util_libs/db/src/schemas/user.rs
+++ b/rust/util_libs/db/src/schemas/user.rs
@@ -38,18 +38,16 @@ pub enum UserRole {
 
 /// Information about a user's role (hoster or developer) in the system
 #[derive(Serialize, Deserialize, Clone, Debug)]
-pub struct UserPubKey {
-    pub pubkey: PubKey,
-    pub is_developer: bool,
-    pub is_hoster: bool,
+pub enum UserPubKey {
+    Developer(PubKey),
+    Hoster(PubKey),
 }
 
 /// User document schema representing a user in the system
 #[derive(Serialize, Deserialize, Clone, Debug, Default)]
 pub struct User {
     /// MongoDB ObjectId of the user document
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub _id: Option<ObjectId>,
+    pub _id: ObjectId,
     /// Common metadata fields
     pub metadata: Metadata,
     /// List of permissions the user has been granted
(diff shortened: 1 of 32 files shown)


rust/util_libs/db/src/schemas/workload.rs
```diff
@@ -36,7 +36,9 @@ pub struct ExecutionPolicy {
 
 #[derive(Serialize, Deserialize, Debug, Clone, Default)]
 pub struct Workload {
-    pub _id: ObjectId,
+    /// MongoDB ObjectId of the user document
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub _id: Option<ObjectId>,
```
JettTech: Can we create the ID in advance to adding the workload in the db?  This  will help us to avoid having to make this an option.

rust/util_libs/db/src/schemas/user.rs
```diff
@@ -38,9 +38,10 @@ pub enum UserRole {
 
 /// Information about a user's role (hoster or developer) in the system
 #[derive(Serialize, Deserialize, Clone, Debug)]
-pub struct RoleInfo {
-    pub collection_id: ObjectId,
+pub struct UserPubKey {
     pub pubkey: PubKey,
+    pub is_developer: bool,
+    pub is_hoster: bool,
```
JettTech: Instead of having these two booleans qualify the type of pubkey, could we just ahve a enum type that includes the pubkey itself?

eg:
```
enum UserPubKey {
  Developer(PubKey)
  Hoster(PubKey)
}
```

JettTech: ..also is there a way we can combine the `UserPubKey` with the `UserRole`?

JettTech (commented):
Overall, this is looking good.  Just left a few comments re the schema updates that I'd like to address prior to merging.

rust/util_libs/db/src/schemas/user.rs
zeeshan595: Yea, I was thinking the same :)
For combining with `UserRole`. It is not likely. Because `UserRole` is referencing the permission the user has when communicating with holo public api. So the roles right now would be Admin, Support, User.
Both the developer and hoster would count under user. So I would prefer to keep them separate

JettTech (approved):
LGTM :shipit:

